PR #6, opened 2019-09-28
Bongo Cat!!!!!

Aeroluna:
I was gonna add more, but I got lazy and decided to just push this
- Bongo Cat is finally here! (Click Kiara in the credits menu)
- Tooltips are slightly smaller and will instantly be revealed instead of after a delay.
- Removed _editorOffset, it was an MM only field that CM doesn't need


Caeden117:
Bongo Cat joins the battle.

PR #8, opened 2019-10-03
Replace checks with string.IsNullOrEmpty()

Jackzmc:
This makes the code, according to my predictions, at least, at the very minimum, I'm pretty positively sure, at least for this exact day I am posting this, nearly, pretty close, to a nice, high value of, a percent, which is, like, approximately, 100% or so, at least I think, not sure actually, perhaps, it could also be, 105% more, in the sense of efficiency, which is the state or quality of being efficient.


Caeden117:
Ah yes. I have never thought about using this super convenient function. I thank you for this huge contribution to ChroMapper.

PR #21, opened 2020-01-07
Added Fully Working "Past Notes Grid"

RyanTheTechMan:
https://cdn.discordapp.com/attachments/661811522523234335/663959552680787969/New_Past_Notes_Grid.mp4


Assets/__Scripts/Map/Notes/BeatmapNoteContainer.cs
```diff
@@ -11,7 +11,7 @@ public class BeatmapNoteContainer : BeatmapObjectContainer {
     [SerializeField] SpriteRenderer dotRenderer;
     [SerializeField] MeshRenderer arrowRenderer;
     [SerializeField] SpriteRenderer swingArcRenderer;
-    [SerializeField] NoteAppearanceSO noteAppearance;
+    [SerializeField] private NoteAppearanceSO noteAppearance { get; }
```
RyanTheTechMan: I don't think this was needed in the end. So it can be undone if wanted.

Caeden117:
Seems good however do keep in mind that 6+ lanes as well as precision position and rotation are things you can do in ChroMapper. I'd recommend making this modular where you can add extra lanes to the UI easily by accessing/modifying a class.

PR #49, opened 2020-02-16
UIMode & New Options

RyanTheTechMan:
[Re-Design Options Menu](https://app.gitkraken.com/glo/board/5e1b3f6f0af39d00106c44f0/card/5e1b41a205746d0010cd3687)
[Playing View](https://app.gitkraken.com/glo/board/5e1b3f6f0af39d00106c44f0/card/5e3f8312a326a200108842b0)
[Preview Mode](https://app.gitkraken.com/glo/board/5e1b3f6f0af39d00106c44f0/card/5e1b4d6ce83aa60010abdf96)


RyanTheTechMan:
@Caeden117 Let me know if you want me to make this 1 click mergeable again.

Caeden117:
Finally got the time to look at the new options menu. I've cloned your fork of ChroMapper and tested it as a separate project.

A few issues that strike me:
1. ~~Credits is noticeably absent from this settings menu (And will be necessary if we want to merge this and fix #58)~~ I eventually found it. Weird spot to put it in my opinion.
2. The `Unlock Grid (360 / 90 Mode)` option's tool tip is the same as the `RGB Lighting` option. As well, the Tooltip for `Hit Sounds` is just `Tooltip`.
3. Speaking of tooltips, the font is still Beon for me (Might just be me or you reverted back to that font)
4. The scrollbar for settings doesn't extend completely from top to bottom (Seriously there's a fair bit of space between the top of the scrollbar and the header). As well, it's a bit thin, and given the large amount of space to the right, we can totally make it a little bit wider and easier to actually click on.
5. `Initial Load Batch Size` is displayed incorrectly (The displayed value is `2` when it should be `100`, `1` should be `50`, `3` should be `150`, etc.)
6. Some text settings can be elaborated more now that there's more room for that. For example, `Chroma Only` can be renamed to `Place Only Chroma Events`
7. I'm not sure why the Master node is separated from the rest of the volume sliders, it should probably be grouped with the rest of them.
8. The color choices for the Dropdown (`Hit Sounds` option) is odd and could use a bit of work. For instance, leave the bright blue color for the scrollbar, instead making the highlighted color a brighter gray.
9. The radial progress thingy that shoots out of sliders can get clipped by settings above and below it (For example, setting `Song Speed` to an option like 170% will have the radial progress thing clipped behind the `Note Lane Amount` option)
10. The container for boolean options is not as rounded as the knob inside of them and therefore isn't completely lined up. On lower resolutions its fine, and is admittedly hard to notice in regular use, however it's one of those instances where once you first notice it is mildly annoying and something you cannot unsee.

This is definitely a good start, and once a majority of these are addressed, I'll pull this into the dev branch. Good work so far!

RyanTheTechMan:
Issue 1: I couldn't come up with a good icon for it, so I just had the credits placed there. My Orignal idea was to have the credits in its own tab at the bottom left. If we can think of a good icon, I'll add it.
Issue 2: My B. All fixed.
Issue 3: I thought I had fixed it, I may have accidentally reverted it? All good now! (Teko)
Issue 4: Made the scroll bar **thiccer**. Moved it up too!
Issue 5: Made `Initial Load Batch Size` show correctly. Also made the `Multiple Offset` value show the changes while in the editor.
Issue 6: Renamed.
Issue 7: Added it to be inside.
Issue 8: Good call. Fixed.
Issue 9: Forced ring to be on top (Layer 15)
Issue 10: Moved, fixed, done, bop, boop, beep.

Issue #38, opened 2020-02-16
Paste ring propagation not working


Xavjer:
It just pastes it in the normal LightMode as 'all lights'

Xavjer:
During analyse, i found various related problems:

1. Paste in RingPropagation will Refresh the pasted to be in the Big Rings column, not on the correct ring. Switching with P twice will correct that, so this is just a visual issue
2. Selecting in RingPropagation and moving them (With the arrows) will move the note on the normal light track, breaking the system and afterwards displaying invalid notes in the propagation view.
3. All rings cannot add new notes in the propagation view, adding a note in the light view and switching to propagation will display it correctly

Probably more to come

PR #55, opened 2020-02-17
Feature/fix pastemove propagation

Xavjer:
Fixes the issues described in #38

Should be tested well, I tried my best and did not find any side effects from my changes.

This also allows to Shift the lights in propagation correctly (left/right/up/down) and in normal lights mode as well as obstacles can now be moved up/down, which was not possible

I will open a new ticket, as there are still other issues which must be checked in relation to ring propagation

### DIFF
--- a/Assets/__Scenes/03_Mapper.unity
+++ b/Assets/__Scenes/03_Mapper.unity
@@ -42668,6 +42668,7 @@ MonoBehaviour:
   selectedColor: {r: 0, g: 1, b: 1, a: 1}
   copiedColor: {r: 0, g: 1, b: 0, a: 1}
   tracksManager: {fileID: 25550679}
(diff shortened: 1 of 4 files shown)


Xavjer:
@Caeden117 Fixed it without nulling the _customData

I found out, that even if you do
`json._customData.Remove("_value")`
then,
`json._customData("_value")`
will still return true
and even worse
`json._customData("_value").AsInt`
will return 0, because the parser is using the double parser =>

```
        public virtual int AsInt {
            get { return (int)AsDouble; }
            set { AsDouble = value; }
        }

        public virtual double AsDouble {
            get {
                double v = 0.0;
                if (double.TryParse(Value, out v))
                    return v;
                return 0.0;
            }
            set {
                Value = value.ToString();
            }
        }
```

Issue #56, opened 2020-02-17
Various Ring propagation Issues


Xavjer:
I will probably comment on this issue multiple times, as I think there may still be various problems with the propagation.

Known at the Moment:

Alt+Clicking in propagation mode resets the position (only visually, swiching back and forth set's it back at it's original place)
=> Dragging within ring propagation does not work correctly

Xavjer:
Open Map with existing Event Nodes
Alt+ Drag a node around, then delete it with Ctrl+Shift+MiddleMouse

Go to Ring Propagation (Press P)

=>

NullReferenceException: Object reference not set to an instance of an object
SimpleJSON.JSONLazyCreator.Set (SimpleJSON.JSONNode aVal) (at Assets/Plugins/SimpleJSON.cs:1044)
SimpleJSON.JSONLazyCreator.set_Item (System.String aKey, SimpleJSON.JSONNode value) (at Assets/Plugins/SimpleJSON.cs:1067)
EventPlacement.OnPhysicsRaycast (UnityEngine.RaycastHit hit, UnityEngine.Vector3 transformedPoint) (at Assets/__Scripts/MapEditor/Mapping/PlacementControllers/EventPlacement.cs:57)
PlacementController`3[BO,BOC,BOCC].Update () (at Assets/__Scripts/MapEditor/Mapping/PlacementControllers/PlacementController.cs:130)

PR #68, opened 2020-02-19
#fixed Dragging ring propagations #56

Xavjer:


Xavjer:
@Caeden117 I needed to know if I am in ring propagation within the BeatmapEventContainer, so I created a method within the TracksManager for it.

I don't know the best practice to do that, maybe just a get method would have been better, if so just tell me and i'll change it accordingly

Caeden117:
I dislike using `TracksManager` to grab Ring Propagation, since the job of `TracksManager` is completely unrelated to that. What I'd recommend is adding a `EventsContainer` parameter into `BeatmapEventContainer.SpawnEvent()`, something like this:

```cs
public static BeatmapEventContainer SpawnEvent(this EventsContainer container, MapEvent data, ...```

Xavjer:
Ok, so I am gonna change that like you suggested this evening.

Xavjer:
@Caeden117 Made the changes as you suggested

PR #80, opened 2020-03-01
Feature/native custom platforms

Xavjer:
Initial version of custom platforms. Tested about 20 different platforms, they are all looking good.

Notes:
I added the Emission to the LightsManager, so that a gameObject can have a normal texture too, otherwise it would still glow when the ControllableLights Material's alpha is 0.

Not done yet:
Spectrogram => Will be added in a later release
Rotating Lights(Lasers) => Not officially supported yet by CustomPlatforms, will be added as soon as they are fully supported
TrackLanes => Add new trackLanes if ringCount exceeds the 15 currently available, so every ring can be propagated.
Unused LightIDs => These LightId's are currently mapped to BackLights, they will be added as separate Lights (The unused in the TrackLane) in the future so they can be mapped as desired.


Assets/__Scripts/MapEditor/Loading/CustomPlatformsLoader.cs
```diff
@@ -0,0 +1,747 @@
...
+    {
+        Material[] materials = (PrefabStageUtility.GetPrefabStage(renderer.gameObject) == null) ? renderer.sharedMaterials : renderer.materials;
+
+        if (materials.Length >= 1 && materials[0] != null)
+        {
+            if (materials[0] != null && (width >= 0.5f))
+                Array.Resize<Material>(ref materials, materials.Length + 1);
+
+            Material lastMaterial = lightMaterial;
+            for (var i = 0; i < materials.Length; i++)
```
Caeden117: I'm pretty sure the code is somewhere around here, however make sure that `ControllableLight` is the only material on lights. Lights on the Camellia Platform will turn red when the controllable lights are off since their default material still exists

Assets/__Scripts/Platforms/Track Rings/TrackLaneRingsManager.cs
```diff
@@ -4,16 +4,17 @@ public class TrackLaneRingsManager : MonoBehaviour
 {
     public TrackLaneRing[] rings { get; private set; }
 
-    [SerializeField] private int ringCount = 10;
-    [SerializeField] private TrackLaneRing prefab;
-    [SerializeField] private float minPositionStep = 1;
-    [SerializeField] private float maxPositionStep = 2;
-    [SerializeField] private float moveSpeed = 1;
+    [SerializeField] public int ringCount = 10;
```
Caeden117: If you are going to change these to public members, please remove `[SerializeField]` since Unity automatically serializes public fields, and change the names of these variables to match `PascalCase`.

Assets/__Scripts/BeatSaberSong.cs
```diff
@@ -300,6 +306,8 @@ public static BeatSaberSong GetSongFromFolder(string directory)
                     case "_coverImageFilename": song.coverImageFilename = node.Value; break;
                     case "_songFilename": song.songFilename = node.Value; break;
                     case "_environmentName": song.environmentName = node.Value; break;
+                    case "_customEnvironmentName": song.customEnvironmentName = node.Value; break;
+                    case "_PlatformName": song.platformName = node.Value; break;
```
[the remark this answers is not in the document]
Xavjer: I am aware of the fact, that CustomPlatforms currently only allows to load one custom environment only and that this behaviour is currently not supported by them.
But I saw, that a lot of the CustomPlatforms currently out there are actually just a platform and nothing else, so I splitted these into an own group and hope, that CustomPlatforms will do the same in the future, so we can have a custom environment with a custom platform in it.

Within CM, this should persist correctly and be stored in the json and also loaded correctly during the song selection menu.

Assets/__Scripts/MapEditor/Loading/LoadInitialMap.cs
```diff
@@ -40,15 +43,19 @@ public IEnumerator LoadMap()
...
         environmentID = SongInfoEditUI.GetEnvironmentIDFromString(song.environmentName); //Grab platform by name (Official or Custom)
-        if (song.customData != null && song.customData["_customEnvironment"] != null && song.customData["_customEnvironment"].Value != "")
+        if (song.customData != null && ((song.customData["_customEnvironment"] != null && song.customData["_customEnvironment"].Value != "") || (song.customData["_customPlatform"] != null && song.customData["_customPlatform"].Value != "")))
         {
-            if (SongInfoEditUI.GetCustomPlatformsIndexFromString(song.customData["_customEnvironment"]) >= 0) {
-                environmentID = SongInfoEditUI.GetCustomPlatformsIndexFromString(song.customData["_customEnvironment"]);
+            if (CustomPlatformsLoader.Instance.GetAllEnvironmentIds().IndexOf(song.customData["_customEnvironment"] ?? "") >= 0) {
+                customPlat = true;
+            }
+            if (CustomPlatformsLoader.Instance.GetAllEnvironmentIds().IndexOf(song.customData["_customPlatform"] ?? "") >= 0)
```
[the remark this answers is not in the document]
Xavjer: See the comment above, if you want, I will remove the customPlatform as it is not supported yet in beat saber(SongCore)

Assets/__Scripts/MapEditor/Loading/CustomPlatformsLoader.cs
Xavjer: From what I saw by various levels, I can't just replace the material.
For example in this level: https://modelsaber.com/Platforms/?id=1565632702&pc
If I remove the GlowWhite, every glowable part would be invisible(transparent) when not active, this would look very strange and is not the behaviour you have in beat saber, where they stay white while not on.

I know that there are maps where this looks strange, but from what I can tell this is as it should be, you normally wouldn't see it (Camellia for example) as it is dark in beat saber and you therefore don't see the color of it.

Assets/__Scripts/Platforms/Track Rings/TrackLaneRingsManager.cs
Xavjer: Will do so

Assets/__Scripts/Platforms/LightsManager.cs
```diff
@@ -168,8 +170,30 @@ private void UpdateColor(Color color, bool emissive, List<LightingEvent> filtere
...
+            foreach (LightingEvent e in filteredEvents)
+            {
+                e.LightMaterial.SetColor(emissive ? "_EmissionColor" : "_BaseColor", color);
+                SetEmission(e.gameObject, (color.a > 0));
+            }
+    }
+
+    private void SetEmission(GameObject gameObject, bool enabled)
+    {
+        Renderer renderer = gameObject.GetComponent<Renderer>();
```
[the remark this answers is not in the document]
Xavjer: Will do so

Assets/__Scripts/MapEditor/Loading/CustomPlatformsLoader.cs
Xavjer: https://imgur.com/a/VTDmbZX
As you can see here, the Lines and the Rings are still visible, even tough they are not enlightened. Maybe I should replace the _replace_X_glow with black (Except the _replace_white_glow), that would probably do the trick. What do you think?

Caeden117: As long as we can get it close to how it looks in Beat Saber, then I'll be happy. I didn't like how the red of the original material stood out, and would definitely be something people would falsely flag as a bug

Xavjer: Okay, I'll therefore do that then, I think this will do the trick

Assets/__Scripts/BeatSaberSong.cs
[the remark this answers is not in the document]
Xavjer: I smoothed the SongEditMenu scene. The only thing left now is your decision, if you want to have the platform removed as this is not yet developed for beat saber itself.

RyanTheTechMan:
I guess this is a good time to start working on a remake of the SongEditMenu?

Xavjer:
That's probably a good idea but probably not as a part of this PR

Caeden117:
Alright I've spent a while to think about it and I've made up my mind.

Since it is not an feature of Custom Platform's, ask Uialeth on Discord to add the separate platforms feature to Custom Platforms. If he has no interest or takes too long to respond then please comment out or remove the `_customPlatform` field and all code relating to it.

Leaving it here with no corresponding feature in the Beat Saber mod will just be confusing to a majority of people. I am 98% sure people would be assuming that it is a feature that also exists in Custom Platforms (which it isn't), and will most likely be creating false bug reports if I pull it like this.

Xavjer:
Okay, for the meantime, I will revert the changes I made for `_customPlatform` this evening and make an own branch for this. If this will be added in CustomPlatforms in the future, I will submit a separate PR that.

I absolutely agree, that it would confuse if this feature is in the map editor but afterwards not reflected within beat saber.

Xavjer:
All necessary parts have been reverted, so now there is only a CustomEnvironment.

If everything goes well, I will have my next PR ready tomorrow, which will contain dynamic EventTrackLanes (Propagation will be as long as there are rings available and the normal view will have the unused removed or replaced/added ones, if the CustomPlatform uses other Id's as triggers)

PR #81, opened 2020-03-06
Made stuff not be in the middle of the screen + transparent thingies?

Futuremappermydud:


Futuremappermydud:
(mainly just a small commit so i can use it without feeling bad)
also i made the page number in song list be teko magenta

Futuremappermydud:
fixed yayyy

PR #84, opened 2020-03-06
Fix CustomPlatforms .plat added by .gitignore

Xavjer:


Caeden117:
This has already been resolved in commit 9facfdee6467238f4969c39b20fed120126d87ca by using AssetBundles instead of AssetDatabases, which among other fixes, keeps the platform files in `Beat Saber/CustomPlatforms` instead of copying them to the *project folder*.

PR #85, opened 2020-03-06
Dynamic event grid size, custom lightsId support and correct hash of customPlatforms

Xavjer:


Xavjer:
Resolved merge conflicts

Assets/__Scripts/MapEditor/Grid/Collections/EventsContainer.cs
```diff
@@ -18,13 +18,24 @@ public bool RingPropagationEditing
...
+            eventPlacement.SetGridSize(value ? (pD.BigRingManager?.rings.Length ?? 15) + 1 : 6 + pD.LightingManagers.Count(s => s != null));
+
+            //ringPropagationLabels.SetActive(value);
+            //eventGridLabels.SetActive(!value);
             UpdateRingPropagationMode();
         }
     }
     private bool ringPropagationEditing = false;
 
+    public void StartUp()
```
Caeden117: a `Start` function exists that is called on Every MonoBehaviour object. Please use that instead.

Assets/__Scripts/Map/Events/BeatmapEventContainer.cs
```diff
@@ -29,6 +29,13 @@ protected override void Awake()
         base.Awake();
     }
 
+
+    protected void Start()
+    {
+        if (eventsContainer != null)
```
Caeden117: See comment on `EventsContainer.cs`

Assets/__Scripts/MapEditor/Grid/Collections/EventsContainer.cs
```diff
@@ -18,13 +18,24 @@ public bool RingPropagationEditing
         set
         {
             ringPropagationEditing = value;
-            ringPropagationLabels.SetActive(value);
-            eventGridLabels.SetActive(!value);
+            PlatformDescriptor pD = FindObjectOfType<PlatformDescriptor>();
+            eventGridLabels.transform.parent.GetComponent<CreateEventTypeLabels>().UpdateLabels(value, value ? (pD.BigRingManager?.rings.Length ?? 15)+1 : 16);
```
Caeden117: Make a `CreateEventTypeLabels` reference and assign it in the Unity Editor instead of grabbing the component manually:

```cs
[SerializeField] private CreateEventTypeLabels labels;
```

Assets/__Scripts/MapEditor/Grid/CreateEventTypeLabels.cs
```diff
@@ -5,59 +5,112 @@ public class CreateEventTypeLabels : MonoBehaviour {
...
+                            break;
+                        case 15:
+                            textMesh.text = "Rotation (Exclude Current Time)";
+                            textMesh.font = UtilityAsset;
+                            break;
+                        default:
+                            if (LightingManagers.Length > i)
+                            {
+                                LightsManager customLight = LightingManagers[i];
+                                textMesh.text = customLight?.gameObject.name;
```
Caeden117: You should be able to get the same name effect by using `customLight.name`

Assets/__Scripts/MapEditor/Grid/CreateEventTypeLabels.cs
```diff
@@ -5,59 +5,112 @@ public class CreateEventTypeLabels : MonoBehaviour {
...
+                    switch (i)
+                    {
+                        case 0:
+                        case 1:
+                        case 2:
+                        case 3:
+                        case 4:
+                            LightsManager e = LightingManagers[i];
+                            textMesh.text = e?.gameObject.name;
+                            textMesh.font = AvailableAsset;
```
Caeden117: Cases 0-4 shouldn't be necessary, the `default` case should work just fine.

Assets/__Scripts/MapEditor/Grid/Collections/EventsContainer.cs
Xavjer: During Start, the PlatformDescriptor is not available (null), I now added a WaitForSeconds(0.1f) like it was done in other scripts, it seems to work. I only hope that this will work every time

Caeden117: If your intention is to get the PlatformDescriptor that has spawned, then look into the `LoadInitialMap.PlatformLoadedEvent` event. You'll need to subscribe to it in the Start function and unsubscribe from the OnDestroy function.

Xavjer: I changed as you suggested (With `PlatformLoadedEvent`) and resolved the merge conflict, ready to be pulled :)

Assets/__Scripts/MapEditor/Grid/Collections/EventsContainer.cs
```diff
@@ -18,13 +19,33 @@ public bool RingPropagationEditing
         set
         {
             ringPropagationEditing = value;
-            ringPropagationLabels.SetActive(value);
-            eventGridLabels.SetActive(!value);
+            PlatformDescriptor pD = FindObjectOfType<PlatformDescriptor>();
```
Caeden117: Making use of the previous `LoadInitialMap.PlatformLoadedEvent`, you can store a local reference of the `PlatformDescriptor` and use that instead of calling the expensive `FindObjectOfType<>()` every time Ring Propagation is changed.

Assets/__Scripts/MapEditor/Grid/Collections/EventsContainer.cs
```diff
@@ -18,13 +19,33 @@ public bool RingPropagationEditing
...
+            labels.UpdateLabels(value, value ? (pD.BigRingManager?.rings.Length ?? 15)+1 : 16);
+            eventPlacement.SetGridSize(value ? (pD.BigRingManager?.rings.Length ?? 15) + 1 : 6 + pD.LightingManagers.Count(s => s != null));
```
Caeden117: For environments without rings (K/DA, Glass Desert, etc.), it would be better to resize the grid so that just "All Rings" is visible.

Pretty much what I'm asking is to make the default size be 0 (before adding 1 for "All Rings").

Caeden117:
I will merge this to the `dev` branch once Pull Request #88 has also been merged, along with other Prs that are open, then merge all of this to `master`

PR #94, opened 2020-04-16
Guesses Beat Saber installation directory for Steam installations.

udoheld:


Assets/__Scripts/UI/FirstBootMenu.cs
```diff
@@ -61,4 +68,25 @@ public class FirstBootMenu : MonoBehaviour {
         helpPanel.SetActive(!helpPanel.activeSelf);
     }
 
+
+    private string guessBeatSaberInstallationDirectory()
+    {
+        //TODO: find out potentiell Oculus Store installation location
```
Caeden117, line 74: You can borrow code from [ModAssistant](https://github.com/Assistant/ModAssistant/blob/master/ModAssistant/Classes/Utils.cs#L253) to find the install location for the Oculus Store version.

udoheld:
You mentioned that you could test the Oculus part with your Oculus Store installation. Please let me know if it works out.

PR #95, opened 2020-04-18
Ignoring more generated / build data.

udoheld:


Thynix:
It looks like this has changes beyond those in `.gitignore`; is that intended?

udoheld:
Hmm, looks strange probably there were some other merges on the dev branch.

PR #101, opened 2020-04-26
Added basic statistics to song editing menu

udoheld:


udoheld:
I think the SongEditMenu had a larger overhaul since this pull request.

PR #102, opened 2020-05-01
Added very basic plugin interface and loader

monkeymanboy:
Look at PLUGINS.md to see how it is used, it's very simple at the moment.


Caeden117:
Seems good! I'll probably be modifying PLUGINS.md a little bit.

PR #104, opened 2020-05-11
Prevent scrolling the timeline when in the pause or settings menu.

Auros:
At first I thought the settings menu was resetting the timeline to zero but it was just me scrolling down in the menus.


Auros:
image

PR #106, opened 2020-05-18
Fix Grid ZDir shader on MacOS

Top-Cat:
Max shader size is 64KB for DirectX but only 16KB for Metal (and OpenGL but we don't talk about OpenGL). Compiled shader for Windows used to look like:
```
Constant Buffer "$Globals" (65536 bytes) on slot 0 {
  Float _GridThickness at 0
  Float _GridSpacing at 4
  Float _Offset at 8
  Vector4 _GridColour at 16
  Vector4 _BaseColour at 32
  Vector1 _BPMChange_Times[2046] at 64
  Vector1 _BPMChange_BPMs[2046] at 32800
  ScalarInt _BPMChange_Count at 65524
  Float _EditorScale at 65528
}
```
Exactly 64KB!

Thankfully the types on Metal uses fewer bytes (16 bytes for a float is crazy though?):
```
Constant Buffer "$Globals" (16424 bytes) {
  Float _GridThickness at 0
  Float _GridSpacing at 4
  Float _Offset at 8
  Vector4 _GridColour at 16
  Vector4 _BaseColour at 32
  Vector1 _BPMChange_Times[2046] at 48
  Vector1 _BPMChange_BPMs[2046] at 8232
  ScalarInt _BPMChange_Count at 16416
  Float _EditorScale at 16420
}
```
That's more than 16KB, but actually not by much. Weirdly I had to drop the array sizes a lot more than you'd expect (there's 624 bytes unaccounted for) but I'm guessing those are local variables or other constants in the shader 🤷‍♂️ 

RIP anyone with 2000 bpm changes who can't edit their map anymore

### DIFF
--- a/Assets/_Graphics/Shaders/Editor/Grid ZDir.shader
+++ b/Assets/_Graphics/Shaders/Editor/Grid ZDir.shader
@@ -35,8 +35,8 @@ Shader "Grid ZDir" {
 			uniform float4 _GridColour;
 			uniform float4 _BaseColour;
 			uniform float _Rotation;
-			uniform float _BPMChange_Times[2046];
-			uniform float _BPMChange_BPMs[2046];
+			uniform float _BPMChange_Times[1963];
+			uniform float _BPMChange_BPMs[1963];
 			uniform int _BPMChange_Count;
 			uniform float _EditorScale;
 
--- a/Assets/__Scripts/MapEditor/Grid/Collections/BPMChangesContainer.cs
+++ b/Assets/__Scripts/MapEditor/Grid/Collections/BPMChangesContainer.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using UnityEngine;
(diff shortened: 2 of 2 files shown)


Caeden117:
Interesting, did not know that this was a shader size issue.

Would it be possible to reduce the 624 unaccounted bytes by removing local variables found in the vertex and fragment shaders? It'd probably read like crap but that could make more room for that one crazy person who needs 2000+ BPM changes in one map.

If it isn't, no big deal since 2000 BPM changes is unreasonably high, and whoever hits that limit should probably find a different song to map.

Top-Cat:
I'm not really an expert this write up was from research I did after I got it working.

I've spent a fair amount of time looking at the compiled shader and trying to work it out but it seems like it's some internal calculation in the compiler that just leads to it crashing with no information. There's no way to debug the compiled version either to see what's in memory when it's running and I had to get these numbers through trial and error, decreasing them until it compiled.

It may be possible to pass larger arrays in global state instead of instanced, but I figured the number was still high enough that it's unlikely to cause many people problems.

Caeden117:
Yeah, I dont expect anyone to run into the max BPM Change cap.

Looks good!

PR #107, opened 2020-05-19
Fix object selection box

Top-Cat:
Example selecting from the lighting side to the waveform side. The old code sets the boundary starting from the division between the right-side lanes instead when moving in this direction the origin is offset and the total size is increased to include the extra lanes.

I've also removed the `IsSelecting = false;` and moved the `OnPhysicsRaycast` out of the placement event as this triggers before the click event that causes notes to be placed. This prevents new notes being placed when ending selection.

Before:
Before
After:
After


Top-Cat:
Sidenote: selecting notes above the bottom row is difficult with the current interface.

The only way I can see to do it is to use the vertical grid either by placing it behind the notes I want to select or starting behind the notes and ending there. In MMA2 the selection box is always full-height so it's not an issue.

Caeden117:
If you want to go into those solutions, I can leave this PR open. If you'd rather not, just tell me and I'll merge the request.

I am currently in the middle of a pretty major refactoring of ChroMapper (does not affect this PR nor the previous two you've submitted) so at this moment I am not really interested in looking at these solutions myself.

Top-Cat:
Nah, just merge this for now.

PR #109, opened 2020-05-24
Spectogram FFT

Top-Cat:
I've kinda gone down a rabbit hole here... I was trying to look at how BPM changes got handled and I made some test files with and immediately noticed the waveform was way out of sync:
image
You may also notice that this sample ends after 39 beats but... there's spectogram over there!

My test file was a simple square wave that should look like this:
image
but the FFT output was different on every beat. It's limited by the attached AudioSource which I think plays the song in realtime?

Elephant in the room: I've read the comment at the top of AudioManager.cs I guess I picked option 1
The output isn't perfect, but I think it's at least as useful as the previous implementation.

The code here spins up 1-4 background threads which do the FFT calculations and (at least on my machine) they do this 5x faster than the old code which means the code is faster even without threading. It also periodically through the process saves the data required to draw the spectogram. This is not a small file, I convert down the floats so they use 2 bytes instead of 4 but the files are still around 2MB for an average 3/4 minute long song.

The upside though is that once the processing is done once, the spectogram for that same song can be rendered in a couple of seconds.

I've not added any configuration for now (you could want to turn the caching off or change the number of threads) but I figure if you want the spectogram at all you'll be happy for it to be faster.

Here's some before and after screenshots:
image
image
image


Assets/__Scripts/MapEditor/Audio/AudioManager.cs
```diff
@@ -1,96 +1,180 @@
...
+        {
+            Thread FFTThread = new Thread(performFFTThreaded);
+            backgroundThreads.Add(FFTThread);
+            FFTThread.Start();
+        }
+
+        Debug.Log("Background Thread Completed");
+    }
+
+    public void performFFTThreaded()
```
Caeden117, line 124: This looks good, but my biggest issue here is the inconsistent naming conventions. I'd appreciate it if methods and public members use `PascalCase`. This also applies to `WaveformCache.cs`.

Assets/__Scripts/MapEditor/Audio/WaveformCache.cs
```diff
@@ -0,0 +1,155 @@
...
+        float[] bandTemp = new float[AudioManager._bands.Length - 1];
+        Debug.Log("WaveformCache: Reading file");
+
+        if (!File.Exists(fileName))
+        {
+            waveformData.chunksLoaded = 0;
+            return;
+        }
+
+        using (FileStream file = File.OpenRead(fileName))
```
Caeden117, line 57: A thought did cross my mind, and that I do want you to address: What if the audio file is swapped or edited? It'd be a little bit uncommon but I have seen people edit the audio file to adjust the offset, and to also add/remove silent beats from the beginning or end. This would require the waveform cache to essentially be thrown out and be regenerated.

How about maybe adding a way to store and check the hash of the song file? An example of grabbing the hash can be found [here](https://github.com/Caeden117/ChroMapper/blob/dev/Assets/__Scripts/Settings/CustomPlatformSettings.cs#L53).

Top-Cat:
So I've added a 2d spectogram which renders slightly faster than the 3d, but while working on it I also 10x'd the 3d render time. As they both process fairly fast now I've removed all the caching code, I also made it so that chunks near the current view are processed first so if you skip to the end it looks fairly instant too.

I've tried to change the member names to PascalCase, tell me if I've missed any :)

image

PR #112, opened 2020-06-08
Fix background lasers following different pattern than in the game

Top-Cat:
Maybe you were expecting a PR for some UI work? Psych! [It's laser time](https://www.youtube.com/watch?v=NsuZphC9bsw).

The lasers in the background in Nice, Triangle, Crab Rave and Monstercat don't match how they look in game. At some point the background lasers were cut in half and the top sections either moved to the center lighting or were removed.

Also bonus fix for monstercat logo in red.

See before/after/in-game screenshots:


Caeden117:
> Maybe you were expecting a PR for some UI work? Psych!

what is this bullshit

while you're at it can you also do it to default environment please thank

Top-Cat:
That juked me because even though the lasers are cut in half, they are at a different angle to the ones that I changed.

Top-Cat:
Final angle

PR #114, opened 2020-06-12
Song Edit Menu

Top-Cat:
There's a lot going on here and it's been a while since I actually did most of the work. Highlights include:

- "It looks clean af" - CyanSnow
- It tells you if there are unsaved changes when you try and navigate away
- Song preview
- Filename validation for cover/audio files
- Copy and paste difficulties


Assets/__Scripts/BeatSaberSong.cs
```diff
@@ -407,7 +407,8 @@ public static BeatSaberSong GetSongFromFolder(string directory)
                                 beatmap.UpdateName(d["_beatmapFilename"]);
                                 set.difficultyBeatmaps.Add(beatmap);
                             }
-                            set.difficultyBeatmaps = set.difficultyBeatmaps.OrderBy(x => x.difficultyRank).ToList();
+                            // If there are already difficulties ignore duplicates of the same difficulty
+                            set.difficultyBeatmaps = set.difficultyBeatmaps.GroupBy(it => it.difficultyRank).Select(group => group.First()).OrderBy(x => x.difficultyRank).ToList();
```
Caeden117, line 411: I do have a [DistinctBy](https://github.com/Caeden117/ChroMapper/blob/dev/Assets/__Scripts/EnumerableExtensions.cs#L6) extension method, could this be used instead of `.GroupBy().Select()`?

Assets/__Scripts/FileOperationAPIWrapper.cs
```diff
@@ -0,0 +1,179 @@
+﻿using System;
```
Caeden117, line 1: Is this an open source file? If so, please include license info/credits.

Assets/__Scripts/UI/SongEditMenu/DifficultyRow.cs
```diff
@@ -0,0 +1,64 @@
...
+    public Toggle Toggle { get; private set; }
+    public Button Button { get; private set; }
+    public TMP_InputField NameInput { get; private set; }
+    public Button Copy { get; private set; }
+    public Image CopyImage { get; private set; }
+    public Button Save { get; private set; }
+    public Button Revert { get; private set; }
+    public Button Paste { get; private set; }
+
+    public DifficultyRow(Transform obj)
```
Caeden117, line 19: Could we perhaps make a `DifficultyRowContainer` script that simply holds references to all of the things we need to grab in this constructor, and just grab stuff from it? I'm not so sure about all this `GetComponent` and `Find` calls, but if it is not being called often then I guess its fine.

Assets/__Scripts/UI/SongEditMenu/ForwardOnClick.cs
```diff
@@ -0,0 +1,17 @@
...
+
+/// <summary>
+/// Used to send clicks from a text field within a button
+/// This is pretty brittle, but it gets the job done
+/// </summary>
+public class ForwardOnClick : MonoBehaviour, IPointerClickHandler {
+
+    public void OnPointerClick(PointerEventData eventData) {
+        var diff = transform.parent.parent;
+        var diffSelector = transform.parent.parent.parent.GetComponent<DifficultySelect>();
```
Caeden117, line 12: would `transform.GetComponentInParent<DifficultySelect>();` work here? Maybe store a serialized private reference to it?

Assets/__Scripts/FileOperationAPIWrapper.cs
Top-Cat: This is a fair question, as far as I can tell the original version of this was posted here https://stackoverflow.com/questions/3282418/send-a-file-to-the-recycle-bin

Which would imply copyright, however I'm not sure there's anything in this file that doesn't derive from the Microsoft dll (and therefore microsoft's documentation https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ifileoperation-setoperationflags , https://docs.microsoft.com/en-gb/windows/win32/api/shellapi/ns-shellapi-shfileopstructa ) which means that it's not itself a copyright-able work as it *has* to be written this way.

Assets/__Scripts/UI/SongEditMenu/DifficultyRow.cs
Top-Cat: This is only called once per row (so 5 times total) when the Scene loads and then not again.

Assets/__Scripts/FileOperationAPIWrapper.cs
Top-Cat: I can strip out most of the file I think if you'd prefer

Caeden117: I think its fine for the reasons you've already said. If it was me personally, I'd comment the URL to the stackoverflow post but that would be up to you.

Assets/__Scripts/BeatSaberSong.cs
Top-Cat: Also updated

PR #117, opened 2020-06-24
Add green day environments

Top-Cat:
Green day environments imported from game assets. Not sure we need to include all the buildings especially as most of them are impossible to see in game.
The ring prop is almost certainly doesn't have the correct values, but at least it's there and does something.

Speakers and the grenade arm are togglable with (L).

https://streamable.com/gp6xaj


Top-Cat:
I've done most of those changes except the buildings are nearly invisible using BasicBlack so I've moved it to using a ring texture material that's slightly brighter.

PR #119, opened 2020-06-28
Add note block LODs

Top-Cat:
Reduces the quality of the mesh used for notes and events when far away from the camera. There's the original 1k vert, 500 vert and 100 vert models which unity will switch based on camera distance. This means if you load a map with a stupid number of lighting events (Hi Aalto) this will give a fair uplift in render performance. See roughly 30% fps increase here:

Before
After


Assets/__Scripts/Map/Notes/BeatmapNoteContainer.cs
```diff
@@ -124,8 +128,11 @@ public void SetIsPlaying(bool isPlaying)
          * 
          * So hard, in fact, that I've given up trying, and instead moved to storing references shaders and swapping them out.
          */
-        Material material = ModelMaterials[0];
-        material.shader = isPlaying ? transparentShader : opaqueShader;
-        material.SetFloat("_Editor_IsPlaying", isPlaying ? 1 : 0);
+        for (int i = 0; i < 4; i++) // Magic number is bad
```
Caeden117, line 131: Magic number _is_ bad! Have you considered using `ModelMaterials.Count`?

Top-Cat: lol
I remember making this comment to make myself go back and fix this and never did, I think it's more complicated than that but I'll get back to you

Top-Cat:
Ok, I've excluded SpriteRenderer materials being added to ModelMaterials (which means that the note visualiser doesn't become grey) and then looped through the entire array to set the shaders.

PR #124, opened 2020-07-05
Tweak placement snapping

Top-Cat:
To reproduce these:
* Set precision to 1/2
* Move into a beat
* Create a bookmark
* Set precision back to 1/1
* Navigate to bookmark

With my mouse hovering over the grid here this is not where I expect the note to be placed.
Bug

Likewise this seems to be intended behaviour and it feels really weird to me that pointing specifically at the vertical grid moves my placement back to the beat but pointing anywhere else doesn't.
Weird

With this change, notes can be placed in increments of the current precision _from_ the current position and round how you would expect. So either:
Place A
Or
Place B
depending on which your mouse is closer to


udoheld:
@Caeden117 that will probably be solved the next time you merge the dev branch with master

PR #127, opened 2020-07-07
Forgot to add the extra property to the transparent shader

Top-Cat:


Top-Cat:
Also realised bomb placement was broken while working on dragging things around

PR #128, opened 2020-07-07
Alt + right click

Top-Cat:
While holding alt if right clicking instead of left clicking the grid will appear at the position the object is being moved so it's easier to position it in 3d space

Not sure if this should move as much of the elements as it does, but it does the job

Very much doesn't work in 360 maps, but nor does alt + lmb

### DIFF
--- a/Assets/__Scripts/MapEditor/Mapping/PlacementControllers/PlacementController.cs
+++ b/Assets/__Scripts/MapEditor/Mapping/PlacementControllers/PlacementController.cs
@@ -178,14 +178,14 @@ public abstract class PlacementController<BO, BOC, BOCC> : MonoBehaviour, CMInpu
                 if (StartDrag(con))
                 {
                     isDraggingObjectAtTime = true;
-                    noteGridTransform.position = new Vector3(noteGridTransform.position.x, noteGridTransform.position.y, con.gameObject.transform.position.z);
+                    noteGridTransform.localPosition = new Vector3(noteGridTransform.localPosition.x, noteGridTransform.localPosition.y, con.gameObject.transform.position.z);
                 }
 
             }
         }
         else if (context.canceled && isDraggingObjectAtTime && instantiatedContainer != null)
         {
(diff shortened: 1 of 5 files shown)


Assets/__Scripts/MapEditor/Mapping/PlacementControllers/PlacementController.cs
```diff
@@ -150,26 +154,63 @@ public void OnInitiateClickandDrag(InputAction.CallbackContext context)
...
+        if (context.performed && CanClickAndDrag)
+        {
+            Ray dragRay = mainCamera.ScreenPointToRay(mousePosition);
+            if (Physics.Raycast(dragRay, out RaycastHit dragHit, 999f, 1 << 9))
+            {
+                BeatmapObjectContainer con = dragHit.transform.gameObject.GetComponent<BeatmapObjectContainer>();
+                if (StartDrag(con))
+                {
+                    isDraggingObjectAtTime = true;
+                    noteGridTransform.position = new Vector3(noteGridTransform.position.x, noteGridTransform.position.y, con.gameObject.transform.position.z);
```
Caeden117, line 181: Change the `noteGridTransform` reference to the interface scaling offset for each grid (`Rotating/Note Grid/Note Interface Scaling Offset`, and then `Rotating/Event Grid/Event Interface Scaling Offset` for the event grid, etc.), and change its local position instead of global position.

Assets/__Scripts/MapEditor/Mapping/PlacementControllers/PlacementController.cs
```diff
@@ -150,26 +154,63 @@ public void OnInitiateClickandDrag(InputAction.CallbackContext context)
...
+                }
+
+            }
+        }
+        else if (context.canceled && isDraggingObjectAtTime && instantiatedContainer != null)
+        {
+            noteGridTransform.position = new Vector3(noteGridTransform.position.x, noteGridTransform.position.y, 0);
```
Caeden117, line 188: See comment above

PR #129, opened 2020-07-13
Undo bug fixes

monkeymanboy:
When an object was moved it changed the position of the BeatmapObject in the placement action, to fix this I made it a duplicate and made DeleteObject handle copied references. Also removed some redundant code in PlacementController for dragging that was causing FinishDrag to be called twice.


Assets/__Scripts/MapEditor/Mapping/PlacementControllers/PlacementController.cs
```diff
@@ -215,13 +215,6 @@ private void FinishDrag()
 
     protected virtual void Update()
     {
-        if ((isDraggingObject && !Input.GetMouseButton(0)) || (isDraggingObjectAtTime && !Input.GetMouseButton(1)))
```
Top-Cat, line 218: This is not duplicate code, it handles an edge case where the buttons are released outside the window

monkeymanboy: Ah figured it was left behind from the old input system, will have to add a check for that then since it was causing 2 modified actions to be added for the same movement.

Caeden117: Do let me know when you find a way to incorporate both.

monkeymanboy: Fixed it turns out the duplicate drag was because both the note placement controller and bomb placement controller were dragging the same note. Fixed by checking the objects type.

monkeymanboy:
I added a fix for epsilon used in both finding conflicting notes and deleting notes, I would have made this a separate PR but I used it in the delete function so I figured I'd throw it in with this one. The way it was before it was doing +/- 0.33333 for the default value which is much larger than intended.

PR #133, opened 2020-07-26
Set bomb material and switched shader to opaque

monkeymanboy:
Fixes issues with the grid rendering over bombs at certain angles where it shouldn't


Caeden117:
ive tested this for a while and didnt spot any noticeable issues, so i'm just gonna say its good to go

PR #147, opened 2020-08-11
Added Chroma Objects Support

AsoVR:
Added a "Place Chroma Objects" button in the Chroma drop down


AsoVR:
Ahh, okay, will do that

AsoVR:
Hello, i have renamed all Variables and Methods to use 'Chroma' or 'Chroma Objects' and changed the Tooltip to reflect this change

PR #148, opened 2020-08-11
Various fixes

Top-Cat:
- Fixes a missing translation key due to a rogue space in `auto .save`
- Fix two bugs with note snapping see:
  - https://streamable.com/7ftvjq - Notes can flip completely over if one is a dot and one is an arrow and they are indexed in the wrong order
  - Another where placing and then toggling the colour of a dot note would infinitely rotate another dot note. This is mostly fixed by applying the refresh to both original and new when toggling colours but as the `+=` were removed from the rest of this function I don't see why this should keep them, seems to work fine like this.
- Fix having to undo twice when dragging notes and bombs. I don't know if the fact you need to be in the matching placement mode is desirable for bombs/notes/walls etc. But currently both note and wall placement controllers are triggered when either is selected and thus double-creates the action at the end. I've made it so you can't drag them both in either mode anymore but it fixes the undo behaviour.
- Added a load of logic that saves the last song, characteristic and diff that you loaded and pre-selects it when loading the edit menu so that assuming you only have one song on the go it will be easy to get in and out. Not sure how to really extend this sensibly without adding more custom data but this is way better than nothing.

### DIFF
--- a/Assets/__Scripts/Settings/Settings.cs
+++ b/Assets/__Scripts/Settings/Settings.cs
@@ -83,6 +83,10 @@ public class Settings {
     public bool Reset360DisplayOnCompleteTurn = true;
     public string Language = "en";
 
+    public string LastLoadedMap = "";
+    public string LastLoadedChar = "";
+    public string LastLoadedDiff = "";
+
     public static Dictionary<string, FieldInfo> AllFieldInfos = new Dictionary<string, FieldInfo>();
     public static Dictionary<string, object> NonPersistentSettings = new Dictionary<string, object>();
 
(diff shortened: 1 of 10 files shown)


Assets/__Scripts/Settings/Settings.cs
```diff
@@ -83,6 +83,10 @@ public class Settings {
     public bool Reset360DisplayOnCompleteTurn = true;
     public string Language = "en";
 
+    public string LastLoadedMap = "";
+    public string LastLoadedChar = "";
+    public string LastLoadedDiff = "";
```
Caeden117, line 88: Is there any reason why we are saving them to the Config file?

Top-Cat: Matches up with MMA2's "Open Last" function (not actually added a button for that) but with the expectation that you're working on one map at a time you can get back to the part you were working on quickly even if you close the app. I could have written separate persistence but the Settings class is so convenient.

PR #152, opened 2020-08-19
Use note position for conflict check

Top-Cat:
Moved the lineIndex and lineLayer to vector calculation into the note object and then used distance rather than checking the index and layer match to determine if they conflict.

Also fixed bomb placement being half a block off vertically.

### DIFF
--- a/Assets/__Scripts/Map/Notes/BeatmapNote.cs
+++ b/Assets/__Scripts/Map/Notes/BeatmapNote.cs
@@ -1,5 +1,6 @@
 ﻿using SimpleJSON;
 using System;
+using UnityEngine;
 
 [Serializable]
 public class BeatmapNote : BeatmapObject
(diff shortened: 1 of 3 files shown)


Assets/__Scripts/Map/Notes/BeatmapNote.cs
```diff
@@ -66,11 +67,39 @@ public class BeatmapNote : BeatmapObject
...
+        return new Vector2(position, layer);
+    }
+
     protected override bool IsConflictingWithObjectAtSameTime(BeatmapObject other)
     {
         if (other is BeatmapNote note)
         {
-            return _lineIndex == note._lineIndex && _lineLayer == note._lineLayer;
+            // Only down to 1/4 spacing
+            return Vector2.Distance(note.GetPosition(), GetPosition()) < 0.25;
```
Caeden117, line 102: I personally think 0.25 units is too much, since some Noodle Extensions maps (especially those made by cyan) have notes super close together, like so:

super close notes

PR #157, opened 2020-08-25
Linkin Park Environment

Top-Cat:


Assets/__Scripts/Platforms/IRotatingLights.cs
```diff
@@ -0,0 +1,8 @@
+﻿using UnityEngine;
+using SimpleJSON;
+
+public abstract class IRotatingLights : MonoBehaviour {
```
Caeden117, line 4: The `I` prefix is for defining [interfaces](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/), not abstract classes.

Assets/__Scripts/Platforms/Track Rings/ITrackLaneRingsManager.cs
```diff
@@ -0,0 +1,10 @@
+﻿using UnityEngine;
+
+public abstract class ITrackLaneRingsManager : MonoBehaviour
```
Caeden117, line 3: Again, the `I` prefix is for defining interfaces.

For this and `IRotatingLights`, it is perfectly possible to convert to an interface. You would need to make sure that the inheriting classes re-inherit `MonoBehaviour`.

Assets/__Scripts/Platforms/Track Rings/LPRingLaserManager.cs
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    public override void HandlePositionEvent()
+    {
+        
+    }
+
+    public override void HandleRotationEvent(JSONNode customData = null)
+    {
+        lpLasers.ForEach(it => it.SwitchStyle());
```
Caeden117, line 22: The left/right rotating lights are not perfectly reflected across the Z axis:

image

When in game, the left/right rotating lights are always perfectly reflected across the Z axis: [video](https://www.youtube.com/watch?v=1wdqfmk5hwI)

Assets/__Scripts/Platforms/IRotatingLights.cs
Top-Cat: Renamed class

Assets/__Scripts/Platforms/Track Rings/ITrackLaneRingsManager.cs
Top-Cat: Renamed class

Assets/__Scripts/Platforms/Track Rings/LPRingLaserManager.cs
Top-Cat: I re-looked at this and I fixed some places where this could happen. It *is* possible to have them not mirrored in game if you set separate laser speeds: 
image but I'm not sure if them getting re-aligned when ring rotation events happen is a feature or a bug.

```csharp
			_rotationDataL.rotationAngle = _randomStartRotation + _rotationDataL.startRotationAngle;
			_rotationDataL.rotationSpeed = Mathf.Abs(_rotationDataL.rotationSpeed);
			_rotationDataR.rotationAngle = 0f - _randomStartRotation + _rotationDataR.startRotationAngle;
			_rotationDataR.rotationSpeed = 0f - Mathf.Abs(_rotationDataL.rotationSpeed);
```

Seems weird to ignore the previous right laser rotation speed

Top-Cat: (Now updated to match the game though)

PR #159, opened 2020-09-06
Bug fixes

Top-Cat:
Some of the things I've seen from discord

Scrolling was really wonky after bpm changes, but I don't think that was even reported over the fact that scrolling past them didn't even work as it was just trying to jump back to the current one so I've added the special case for when scrolling at a bpm change.

Gridlines using highest prime factor:

image

Fixed boost handling to match the fact you can actually get three colours at once if you make use of inverted lights:
image


Caeden117 (approved):
alright, looks good to me

PR #160, opened 2020-09-14
fixed selection bug

monkeymanboy:
Select between was only looking through loaded containers so large selections would not work. Box select also has a similar issue but this PR does not fix that.


Caeden117 (approved):
Box select as a fix in place which makes large selections possible, but it should probably be rewritten from scratch to not use containers.

Anyways, PR looks good to me

PR #161, opened 2020-09-24
Fixes metronome drift

Splamy:
This is a minor fix for the metronome.
In the current implementation the ticking will drift from the target bmp depending on the fps.
So about 118/120bpm or 150/155bpm at 60fps in my testing.
The change preserves the already elapsed time into the new frame.

I also tried to find a constant high precision timer for unity since the current one still only fires at each frame at max. But it seems unity cannot do that. Unless maybe with a own thread for that, however that seems overkill.


Caeden117 (approved):
Nice catch! Seems like this problem was more minimal for myself thanks to my 144Hz panel.

> I also tried to find a constant high precision timer for unity since the current one still only fires at each frame at max. But it seems unity cannot do that. Unless maybe with a own thread for that, however that seems overkill.

Admittedly, Unity has a whole lot of mini-flaws that I've had to deal with. But I also agree that a high precision timer on a separate thread for the Metronome is a tad bit overkill.

Either way, looks good to me 👍

PR #162, opened 2020-10-03
Swings per second

Top-Cat:
Essentially a copy-paste of the python code, only the total overall SPS is shown in the UI but the values for red, blue and total x Overall, peak and median are calculated

image

Bonus difficulty creation bugfix too


Caeden117 (approved):
*ahem*

k

PR #164, opened 2020-10-24
Camera location hotkeys

Top-Cat:
Probably more code in here for saving object types in the existing settings file than actually adding the feature.

Hotkeys pretty much copied from Starcraft 2.

F5-F8 are the default recall hotkeys.
Ctrl + <Key> = Set current location
And a second layer is accessible by adding shift allowing for 8 locations.

### DIFF
--- a/Assets/__Scripts/Input/Master.cs
+++ b/Assets/__Scripts/Input/Master.cs
@@ -65,6 +65,38 @@ public class @CMInput : IInputActionCollection, IDisposable
                     ""expectedControlType"": """",
                     ""processors"": """",
                     ""interactions"": """"
+                },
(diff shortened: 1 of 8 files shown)


Caeden117 (approved):
`@Rabbit`

I like `IJSONSetting`, but if it's going to get more complex then we might want to switch settings to instead use something like `System.Text.JSON` or `Newtonsoft.JSON` to perform the heavy lifting of dynamically parsing the settings file into the object.

Caeden117:
Added bonus: I've made the camera automatically teleport to the first camera location when the editor scene loads, for anyone who does not like CM's default camera location

PR #165, opened 2020-10-24
Small bugfixes

Top-Cat:
Fixes for:

Notes not ticking when playing the same note twice, see:
https://discordapp.com/channels/702230714585710602/702232127755780197/763963021042253844

Spectrogram moving when pausing / changing UI mode etc. See:
https://discordapp.com/channels/702230714585710602/702232127755780197/767122138742128692


Caeden117 (approved):
k

PR #166, opened 2020-10-25
More bugfixes

Top-Cat:
Shift + up and down won't refresh window-aligned notes
Don't know if the same as:
https://discordapp.com/channels/702230714585710602/702232127755780197/768040651383635969

Alt+Shift+Left click shouldn't move blocks
https://discordapp.com/channels/702230714585710602/702232127755780197/767186327682678825

alt+left click and alt+right click will deselect a note that was previously selected
https://discordapp.com/channels/702230714585710602/702232127755780197/767890235005730847

Fixed some undo bugs by using IsConflictingWith, I don't know how objects are getting confused here but they certainly were. I couldn't work out how to reproduce it every time either...


Caeden117 (approved):
chromapper

more like

buggymapper

PR #167, opened 2020-10-26
Add _speed property, apply new properties before multipliers

Top-Cat:
Pretty self explanatory, helpful link to current chroma code being emulated here:
https://github.com/Aeroluna/Chroma/blob/master/Chroma/HarmonyPatches/TrackLaneRingsRotationEffectSpawner.cs#L112


Caeden117 (approved):
k

PR #170, opened 2020-11-04
Change ring update loop code to match chroma

Top-Cat:
Updated to match chroma
See https://github.com/Aeroluna/Chroma/blob/master/Chroma/ChromaRingsRotationEffect.cs#L64

Fix for wonky rounding:

```        float a = 2.5f;
        for (int i = 0; i < 10; i++)
        {
            Debug.Log(Mathf.RoundToInt(a));
            a++;
        }
```

image


Caeden117 (approved):
why tf do c# be like this

PR #171, opened 2020-11-06
Add release channel dropdown in experimental

Top-Cat:
Watch this space

Obviously does nothing immediately, but I think I need a "stable" build to support this setting before it makes sense to do anything else.

experimental


Caeden117:
how about i accept this PR so you have to keep making new ones

haha... just kidding...

...unless?

Top-Cat:
image

What's that up there?

Caeden117 (approved):
looks good to me

wait i mean "k"

PR #172, opened 2020-11-12
BTS Environment

Top-Cat:
The only code changes here are:

- Adding a multiplier so that the z-axis laser offset can be applied in reverse to sync clockwise and counter-clockwise lasers
- Allowing the first ring to move during a zoom as I've combined all the small pillars into a single object and called it a ring
- Making it possible for a rotation effect to apply to multiple ring managers so that the mirrors can move identically

Everything else is just prefabs :)


Caeden117 (approved):
ladies and gentlemen, may i have your attention please?

k

thank you for coming to my ted talk

PR #174, opened 2020-11-15
Add additional diagonal note keybinds on numpad

Top-Cat:
Simple options binds on the numpad if you don't want to press two keys


Caeden117 (approved):
mmmmmmkay

PR #175, opened 2020-11-15
Add note transparancy via dithering

Top-Cat:
Removes the transparent / opaque switching when going into play mode and replaces with a single shader that uses dithering for transparency.

This shader _does_ support only making half a block transparent across the grid, however instead it listens for track movement and toggles the entire transparency when the centre crosses it.

Before / After:

image


Caeden117 (approved):
didnt see any problems with this when playing through maps, guess it'll please more people who preferred MM's design to CMs 

also k

PR #176, opened 2020-11-15
Ui scale

Top-Cat:
Because adjusting it was a PITA

Depends on my release channels PR as I don't want to deal with merge conflicts from editing the options screen.


Caeden117 (approved):
coolio

PR #177, opened 2020-11-15
Lane refactor

Top-Cat:
Refactor lanes, hold what's being displayed in list in memory and use that to map between co-ordinates and types instead of static array lookups that break when we start hiding lanes.

Hide 360/90 when not in a 360 map so that we don't get random warnings all the time
Fix shifting events off the edge of the world


Caeden117 (approved):
ok mr first time contributor

PR #178, opened 2020-11-16
Bug fixes

Top-Cat:
BTS logo in correct lighting group
Re-enable the note tick object that I accidentally disabled, but less opacity

Bonus optimisations, cache state of a couple of values so we don't give unity work to do if they haven't changed.
PastNotesWorker.OnTimeChanged code is now 4x more lines but at least twice as fast if not more


Top-Cat:
Good thing too as I forgot to add a file before

Caeden117 (approved):
back to normality: k

PR #181, opened 2020-11-17
SongCore overrides fixes

Top-Cat:
Bringing back the PR I closed a while back, but better
Store platform default colours and user options in separate values so we can reset properly when clicking the buttons

Also forced the alpha values to 1 for all these as SongCore refuses to load the values if they don't contain exactly 3 elements 🤔


Caeden117 (approved):
k

PR #182, opened 2020-11-18
Make box select great again

Top-Cat:
Refactor selection box to not use a BoxColider as now that objects are not loaded all the time this has funky results.

Instead I've implemented ways of working out where the centre of most objects are and a not terrible way of working out if they are within the bounding box.

The bounding box is also itself bounded to the track being selected, no more can you move your cursor across tracks only to find the filter doesn't allow that anyway.

The biggest impact here is that, instead of selecting things if any part of them is within the bounding area, things are only selected when over half of them is in. I think this only really affects precision placement. Thoughts welcome.

### DIFF
--- a/Assets/__Scripts/Map/Events/MapEvent.cs
+++ b/Assets/__Scripts/Map/Events/MapEvent.cs
@@ -76,7 +76,7 @@ public class MapEvent : BeatmapObject {
         return null;
     }
 
-    public Vector2? GetPosisition(CreateEventTypeLabels labels, int prop)
+    public Vector2? GetPosition(CreateEventTypeLabels labels, int prop)
     {
         if (prop >= 0)
         {
--- a/Assets/__Scripts/Map/Events/MapEvent.cs
+++ b/Assets/__Scripts/Map/Events/MapEvent.cs
@@ -76,6 +76,41 @@ public class MapEvent : BeatmapObject {
         return null;
     }
 
+    public Vector2? GetPosisition(CreateEventTypeLabels labels, int prop)
+    {
+        if (prop >= 0)
+        {
+            if (_type == prop)
+            {
+                if (_customData != null &&
+                    _customData.Count > 0 &&
+                    _customData.HasKey("_propID")
+                    && _customData["_propID"].IsNumber)
+                {
+                    return new Vector2(
+                        _customData["_propID"] + 1.5f,
+                        0.5f
+                    );
+                }
+                else
+                {
+                    return new Vector2(
+                        0.5f,
+                        0.5f
+                    );
+                }
+            }
+            return null;
+        }
+        else
+        {
+            return new Vector2(
+                labels.EventTypeToLaneId(_type) + 0.5f,
+                0.5f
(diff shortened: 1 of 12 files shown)


Assets/__Scripts/MapEditor/Mapping/Selection/BoxSelectionPlacementController.cs
```diff
@@ -32,21 +32,56 @@ public class BoxSelectionPlacementController : PlacementController<MapEvent, Bea
 
     public override int PlacementXMax => int.MaxValue;
 
+    public Bounds bounds = default;
+    public CustomEventsContainer customCollection;
+    public EventsContainer eventsContainer;
+    public CreateEventTypeLabels labels;
+
+    // And I thought generics would make this method cleaner
+    private void TestForType<T, BO, BOC, BOCC>(RaycastHit hit, BeatmapObject.Type type) where T : PlacementController<BO, BOC, BOCC> where BO : BeatmapObject where BOC : BeatmapObjectContainer where BOCC : BeatmapObjectContainerCollection
```
Caeden117, line 41: If you want to keep the generics clean, what you can do is make a non-generic `PlacementController` class, which would have all of the non-generic methods (if there even is any lmao), then have the generic `PlacementController<BO, BOC, BOCC>` class inherit that.

That, or you can just make the T argument `MonoBehaviour`, which for this case wouldn't even need the other 3 type arguments.

Assets/__Scripts/MapEditor/Mapping/Selection/BoxSelectionPlacementController.cs
```diff
@@ -83,13 +118,55 @@ public override void OnPhysicsRaycast(RaycastHit hit, Vector3 transformedPoint)
...
+
+                var left = instantiatedContainer.transform.localPosition.x + instantiatedContainer.transform.localScale.x;
+                var right = instantiatedContainer.transform.localPosition.x;
+                if (right < left) (left, right) = (right, left);
+
+                var top = instantiatedContainer.transform.localPosition.y + instantiatedContainer.transform.localScale.y;
+                var bottom = instantiatedContainer.transform.localPosition.y;
+                if (top < bottom) (top, bottom) = (bottom, top);
+
+                if (bo is BeatmapNote note)
```
Caeden117, line 137: With all of these Position and Center variables, I almost want to consider abstracting it all into `BeatmapObjectBounds` class, and having an abstract method/property in `BeatmapObject` which constructs it from the map data.

Assets/__Scripts/Map/Events/MapEvent.cs
```diff
@@ -76,6 +76,41 @@ public static bool IsBlueEventFromValue(int _value)
         return null;
     }
 
+    public Vector2? GetPosisition(CreateEventTypeLabels labels, int prop)
```
Caeden117, line 79: > GetPosisition

Nice.

Caeden117 (approved):
mmmk, looks good

PR #183, opened 2020-11-18
Small fix/optimisation

Top-Cat:
Show rotation lanes even outside of 360/90 maps if rotation events exist

Removed some more linq because it is "slow" and called 100s of times when scrolling through the map and cache the result of 1/pow(10, 3) ~0.15ms -> 0.05ms in deep profiling


Caeden117 (approved):
kay

PR #184, opened 2020-11-20
Move contributor editing to SongEditUI, add file browser support

Top-Cat:
File browser library is MIT licenced

Song files, cover images and contributor images can now be added by finding them with a native file browser

Bonus bug fix that stops contributor images being added twice if used in multiple contributors.


Caeden117:
Perhaps move the browse icons down into the text box, and move the existing status icons to the left to fit them? mainly I think it would allow the browse icons to be larger and a tad bit easier to click

image

Top-Cat:
Ok icons now look like this
image

Caeden117 (approved):
alright, this all looks pretty good!

PR #185, opened 2020-11-20
Bug fixes

Top-Cat:
Move grids down to align with when the vertical grids are

Before:
image
After:
image

Changed LoadInitialMap code, I kinda assume it was always broken but didn't show up until I changed the order things loaded a bit

Moved point where bocks become transparent to hopefully guard against floating point precision errors


Top-Cat:
Fixed some bugs that caused reloads to corrupt when using propIDs

Fixed shift+click and drag across events not working due to 🔺 changes

Caeden117 (approved):
aight

PR #186, opened 2020-11-21
Fix creating past notes forever if no new ones pass the cursor

Top-Cat:
Major fps drop when scrolling through a map with no notes :(


Top-Cat:
Also as Jamman360 can attest I used a drill to save scrolling in the process of reproducing this bug

Caeden117 (approved):
eee

PR #189, opened 2020-11-21
Prop shift optimisations

Top-Cat:
So first we call UpdateGridPosition, then set PropagationEditing which updates the locations of all the objects and then we call RefreshPool(true) which recycles all the containers we just made sure were in the right places. Seems efficient.

This is worst when in prop view as the locations of all the events are updated _for each event being moved_

Moved to a plinq which may be slower for small number of events, but fast enough for action from user input.

Also cut down on calls to generate copy, as after this recent commit two calls here are superfluous https://github.com/Caeden117/ChroMapper/commit/0a177c1c3abb1045efdc783be3d648f1fa8774e1


Caeden117 (approved):
> caeden

> working editor

> performant code

pick two

PR #190, opened 2020-11-21
Bug fix train has arrived

Top-Cat:
Sort the bookmarks array, fixes weird next/previous behaviour see https://discordapp.com/channels/702230714585710602/702232127755780197/779680425391292426

Harden up CMInputCallbackInstaller to stop rogue classes locking out the user, each type can only have a single block of an input type. Fixes editing colour selection as that called disable for each keystroke see https://discordapp.com/channels/702230714585710602/702232127755780197/779682737917001758

Re-enable escape and tab on CM_Inputbox and CM_Dialogbox

Fix random blocks changing direction as the dragged container is never cleared so when it gets reused somewhere it will get updated when you press directions as if you were still dragging see https://discordapp.com/channels/702230714585710602/702232127755780197/779738549251080203

Don't allow shift + click to select unloaded objects, especially if they're the queued object, we update the properties of this object a lot and it gets weird copy/pasting with it see https://discordapp.com/channels/702230714585710602/702232127755780197/779735929677414410

### DIFF
--- a/Assets/HSVPicker/UI/HexColorField.cs
+++ b/Assets/HSVPicker/UI/HexColorField.cs
@@ -31,7 +31,7 @@ public class HexColorField : MonoBehaviour
     private void TextValueChanged(string arg0)
     {
         if (!hexInputField.isFocused) return;
-        CMInputCallbackInstaller.DisableActionMaps(typeof(CMInput).GetNestedTypes().Where(x => x.IsInterface));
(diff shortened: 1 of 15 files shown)


Caeden117 (approved):
my only nitpick is explicitly using types, like `typeof(HexColorField)`, when you could use `GetType()`, *or* use `StackTrace` inside the method to yoink the declaring class without it needing to explicitly pass itself, but again those are all nitpicks and I guess i can live with it.

PR #191, opened 2020-11-21
Split prop ids when running strobe gen

Top-Cat:
Moved prop id splitting up to the main function as otherwise we delete all other prop notes nearby

Also fixes prop notes not being visible if spawned while propagation editing is enabled but not their event type

### DIFF
--- a/Assets/__Scripts/Map/BeatmapObject.cs
+++ b/Assets/__Scripts/Map/BeatmapObject.cs
@@ -1,6 +1,5 @@
 ﻿using SimpleJSON;
 using System;
-using System.Security.Cryptography.X509Certificates;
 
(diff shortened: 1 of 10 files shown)


Caeden117 (changes requested):
Main thing here:

Since you've moved the grouping of prop IDs to further up the chain, we should probably modify the abstract method of `StrobeGeneratorPass` to take in a `int? propId` just like we are doing for the Lighting and Step Gradient passes. That'll also allow us to not have to re-check whether or not we have ring prop further down the chain.

Top-Cat:
Of course I realise that lets us delete a load of code trying to split them out when we can only have one or the other

Caeden117 (approved):
okay, now this looks good

PR #192, opened 2020-11-22
Make transparency cull dependent on user precision and editor scale

Top-Cat:
Also undid accidental change to raycaster


Assets/__Scripts/MapEditor/Grid/Collections/NotesContainer.cs
```diff
@@ -12,16 +12,19 @@ public class NotesContainer : BeatmapObjectContainerCollection {
 
     private HashSet<Material> allNoteRenderers = new HashSet<Material>();
     private float epsilon = 0.001f;
+    public static float translucentCull = -0.001f;
```
Caeden117, line 15: this hurts me and the .NET naming conventions

> DO use PascalCasing for all public member, type, and namespace names consisting of multiple words.

Top-Cat: image

Caeden117 (approved):
kk ty

PR #193, opened 2020-11-22
Multi-node editor

Top-Cat:
Attempts to merge object json and then apply patches back to the original objects. Code is a little ugly but 🤷‍♂️ 

Only shared fields are displayed, although you can update anything. Shared fields with differing values are shown with null values.

Example below, obviously this is already possible with ctrl+arrows

image


Caeden117:
This is just a style nitpick, but is it possible to show something like `-` instead of `null` for JSON fields that have different values? Something akin to Unity multi-editing

image

Top-Cat:
> This is just a style nitpick, but is it possible to show something like `-` instead of `null` for JSON fields that have different values? Something akin to Unity multi-editing
> 
> image

Not without re-architecting it, it uses the JSON library throughout and so it has to still be valid json. Closest I could get is replacing those values with a quoted dash. "-"

Top-Cat:
Or I guess we could mess with the JSON library to think that dashes are valid json

Top-Cat:
Turns out SimpleJSON already reads strings regardless of if they're quoted or not so {"_time": -} is already considered to be valid (even though it's not). So I added a json class that writes an unquoted dash to the stringbuilder and just check if the json string from the parser is a dash.

I hope dash is never actually a string anyone wants to use

Caeden117 (approved):
Alright, this works great

PR #194, opened 2020-11-25
Remove OSC code from core

Top-Cat:
Lets be honest, this code doesn't even support chroma 2.0

Anyone who wants to use this (is that anyone?) can add it back as a plugin:
https://github.com/Top-Cat/CM-OSC/releases/tag/0.1.0


Caeden117 (approved):
can we get an f in chat for osc

PR #196, opened 2020-11-26
Node editor customisation & fixes

Top-Cat:
Add text/ui size node editor settings, clean up ui

Also fixed editing off-screen objects (largely by deleting unneeded code)

By default:
image
With settings changed by user:
image


Caeden117 (approved):
i have nothing to say so E

PR #197, opened 2020-11-27
Fix notes not deleting properly after being moved in time

Top-Cat:
So after hours of confusion and actually stepping through the source code for SortedSet<T> I learned that for delete to function as expected the items must still be sorted in the same way as when they were added to the set. This means any time we want to change the _time property of an object we need to re-add it to the set otherwise there's a chance when the tree is traversed later it won't work.

### Demo code:
https://ideone.com/wpLvu3

### How to reproduce in CM:
On a new map place a single block
Place a second block later in time
Drag or shift move the second block before the first in time
Delete the second block
Save and exit
Reload the map
Observe the second block has not been deleted

----

At the time time I realised that if you move a note to a precision where _time will have more decimal places than the current selected precision then the undo will fail as the _time will not match the rounded _time saved in the BeatmapAction.

To fix this we need to check within the current precision for deletes (this should still be 15x smaller than it's possible to place blocks together at 3dp, it will fall over at 1dp but come on)

### DIFF
--- a/Assets/__Scripts/MapEditor/Grid/Collections/NotesContainer.cs
+++ b/Assets/__Scripts/MapEditor/Grid/Collections/NotesContainer.cs
@@ -11,7 +11,7 @@ public class NotesContainer : BeatmapObjectContainerCollection {
     [SerializeField] private TracksManager tracksManager;
 
     private HashSet<Material> allNoteRenderers = new HashSet<Material>();
-    private float epsilon = 0.001f;
+    public static float Epsilon = 0.001f;
     public static float TranslucentCull = -0.001f;
 
     private void Start()
@@ -23,8 +23,8 @@ public class NotesContainer : BeatmapObjectContainerCollection {
 
     private void UpdateEpsilon(object precision)
     {
-        epsilon = 1 / Mathf.Pow(10, (int)Settings.Instance.TimeValueDecimalPrecision);
-        TranslucentCull = -Settings.Instance.EditorScale * epsilon;
+        Epsilon = 1 / Mathf.Pow(10, Settings.Instance.TimeValueDecimalPrecision);
+        TranslucentCull = -Settings.Instance.EditorScale * Epsilon;
     }
 
     public static bool ShowArcVisualizer { get; private set; } = false;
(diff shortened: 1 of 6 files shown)


Assets/__Scripts/MapEditor/Grid/Collections/NotesContainer.cs
```diff
@@ -11,7 +11,7 @@ public class NotesContainer : BeatmapObjectContainerCollection {
     [SerializeField] private TracksManager tracksManager;
 
     private HashSet<Material> allNoteRenderers = new HashSet<Material>();
-    private float epsilon = 0.001f;
+    public static float Epsilon = 0.001f;
```
Caeden117, line 14: Because `BeatmapObject` is now using this, I think it'll make sense to move this to `BeatmapObjectContainerCollection`. Hopefully nothing should break from that, since this class inherits that anyways.

Assets/__Scripts/MapEditor/Mapping/PlacementControllers/PlacementController.cs
```diff
@@ -216,10 +216,16 @@ private void FinishDrag()
         if (!(isDraggingObject || isDraggingObjectAtTime)) return;
         //First, find and delete anything that's overlapping our dragged object.
         var selected = SelectionController.IsObjectSelected(draggedObjectData);
-        objectContainerCollection.RemoveConflictingObjects(new[] { draggedObjectData }, out List<BeatmapObject> conflicting);
+
+        // To delete properly we need to set the original time
+        float _time = draggedObjectData._time;
+        draggedObjectData._time = originalDraggedObjectData._time;
+        objectContainerCollection.DeleteObject(draggedObjectData, false, false);
```
Caeden117, line 223: Couldn't we just delete `originalDraggedObjectData`?

Top-Cat: No, the draggedObjectData points to an object that is inside LoadedObjects, it needs to have the correct _time value when it's removed.

You could _call_ this function with originalDraggedObjectData but you still have to make the change to draggedObjectData or it won't be removed.

Caeden117 (approved):
sorted collections are wack

PR #198, opened 2020-11-27
Strobe gen go brrrr

Top-Cat:
Couple of performance optimisations for copying events and notes as they are most common to spam

Big boost in strobe gen on high event count maps by using the sorted set to exclude objects before and after the selected region.


Caeden117 (approved):
cm performance

stonks

PR #199, opened 2020-11-29
Tests?

Top-Cat:
Added some tests covering a few of the recent bugs I've fixed. Mostly want to start reducing the possibility of re-introducing old issues when fixing new ones.

Few places where stuff is too tightly coupled currently to be able to test it (dragging has a load of raycasting for instance). But hopefully in time that can improve.

I've had to split up the code into separate dlls so that the tests aren't packaged into the build. Also code that can only be run in the editor can't be referenced by test builds so that all moved under the Editor folder.

Swapped out the official `com.unity.test-framework` package for a github version that adds UnityOneTimeSetUp which accepts an IEnumerator type method so that we can spend time loading the mapper scene before the tests start

image


Caeden117 (approved):
wait fuck this means i have to unit test my stuff

PR #200, opened 2020-11-30
Change default release channel to stable

Top-Cat:
Probably don't merge this yet?

I'll promote whatever build is created after this is merged to be the current "stable" build in the launcher, so lets make it a good one. Very much focused on bug fixing atm so it may not have to be that long.


Caeden117:
> I'll promote whatever build is created after this is merged to be the current "stable" build in the launcher, so lets make it a good one.

bold of you to assume that chromapper will get to that point

Top-Cat:
Honestly just needs to be as good/better than 242 and given the commit message for that was "Temporarily swap precision placement to Shift + Alt + Click" I think we're good.

PR #201, opened 2020-11-30
Prop fixes

Top-Cat:
I think this fixes the rng involved in loading the green day environment.

Also fixed ctrl+arrows moving prop events around. For now I've re-enabled the ability to move off into empty space to the right of prop ids, but this isn't hard to disable if we want to stop users doing it. There was a small discussion on this point here:
https://discordapp.com/channels/702230714585710602/702232127755780197/780418721311096923


Caeden117 (approved):
k

PR #202, opened 2020-11-30
Bpm block fixes

Top-Cat:
Fix not being able to click on the block through the text. Tell me if there was a reason this raycaster was here?

Fixed metronome not changing when bpm block was passed


Caeden117 (approved):
raycaster was probably autogenerated by unity because canvas reasons

PR #203, opened 2020-12-01
Slightly squashed model to reduce z-fighting

Top-Cat:
Reduce event z-fighting by slightly pushing the back together. Some pyramid for everyone

It's subtle enough that you still get some z-fighting but only when events are very close together and viewed from a distance
image


Caeden117 (approved):
Sure, but I'm not sure if you can truly fix z-fighting unless you submit to the pyramid overlords

PR #204, opened 2020-12-03
Manually fix prop id mappings

Top-Cat:


Caeden117 (changes requested):
Per discussion on discord: Please abstract group multiplier and offset from `LightingEvent` to `LightsManager`. It would make it much easier to tweak these values, as well as using them in the `CorrectPropIDs` script.

Caeden117 (approved):
Adding other platform elements (Specifically, KDA's... ground spikes?) seemed to be difficult with this system, but I'll accept anyways for the changes on the other environments.

PR #205, opened 2020-12-04
Spruce up the first boot menu, add language selection

Top-Cat:
image

There's some code that somewhat-naively checks system specs and dynamically changes the default settings.

### DIFF
--- a/Assets/__Scripts/UI/FirstBootMenu.cs
+++ b/Assets/__Scripts/UI/FirstBootMenu.cs
@@ -1,4 +1,6 @@
-﻿using System.Globalization;
+﻿using System;
+using System.Collections;
+using System.Globalization;
 using System.Collections.Generic;
 using System.Text.RegularExpressions;
 using UnityEngine;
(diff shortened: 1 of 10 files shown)


Caeden117:
> SteamIsStupid

Top-Cat:
> > SteamIsStupid

Fun fact: Steam is not installed on the K drive

Assets/__Scripts/UI/FirstBootMenu.cs
```diff
@@ -48,38 +55,86 @@ public class FirstBootMenu : MonoBehaviour {
...
+
+    public void SetDefaults()
+    {
+        // Auto scale settings for machine
+        if (SystemInfo.processorCount <= 2)
+        {
+            Settings.Instance.Waveform = 0;
+        }
+
+        if (SystemInfo.graphicsMemorySize <= 1024)
```
Caeden117, line 94: Is it wise to determine performance just by the amount of graphics memory the user has? This could be better off with a simple `Graphics` dropdown with basic options `Performance`, `Balance`, and `Quality`

Top-Cat: I've still left the gpu memory -> initial selected value, but you can now see it's happening and choose a different value

image

Caeden117 (approved):
Alright, much better

PR #206, opened 2020-12-05
added basic theme system and redid song select screen

monkeymanboy:
More PRs will follow with new options screen, song edit screen, and in editor UI.


Top-Cat:
image

First impressions:
- Feels entirely separate to the other existing menus, I know this was somewhat true when I added the current song edit screen but that at least feels _close_ to the settings screen. I know you've set your goal to replace them all (and I'm obviously biased) but I prefer the look of the song edit screen to this screen.
- I like being able to see more information, the images are nice. But people have suggested showing the folder names too. Maybe other song metadata if we can?
- Don't like the placement of the new map button, at least at its size
- Feedback from the refresh icon is very hard to see

BrightKnight02:
I gotta agree with top_cat on the new map button. It seems a little out of place.

Top-Cat:
We're also missing a translation metadata file which is why they're broken in that screenshot

monkeymanboy:
> * Feels entirely separate to the other existing menus, I know this was somewhat true when I added the current song edit screen but that at least feels _close_ to the settings screen. I know you've set your goal to replace them all (and I'm obviously biased) but I prefer the look of the song edit screen to this screen. Maybe partly as we've lost the chroma.
All the menus are replaced on another fork in a style similar to this, including the settings screen. They just need to be updated to support the theming system.
> * I like being able to see more information, the images are nice. But people have suggested showing the folder names too. Maybe other song metadata if we can?
Maybe there can be a little info icon on the right with a hover hint that has more info?
> * Don't like the placement of the new map button, at least at its size. Why not on the left?
Having it on the left would cover the name and icon of the maps at the bottom of the screen.
> * Feedback from the refresh icon is very hard to see
Noted, will make a darker color.
> * Moving between WIP / Normal takes a second, would be nice to optimise at least hiding the old list to make it feel snappier
Yeah, I do wanna improve this a bit. They don't need to be fully loaded as BeatSaberSongs but they are at the moment which I suspect is a big source of the stutter.

And yeah looks like I forgot to push the translation file my bad.

Top-Cat:
> Having it on the left would cover the name and icon of the maps at the bottom of the screen.
You're not thinking left enough

BrightKnight02:
what if the new map button was where the search and refresh buttons are and those two icons were moved a bit to the left?

Top-Cat:
> All the menus are replaced on another fork in a style similar to this, including the settings screen. They just need to be updated to support the theming system.

If you want buy in on that you may want to post some screenshots / talk about it in the discord? Just saying "Well I've done it" isn't always the best salespitch

monkeymanboy:
Fair enough, here's how the others look
settings
edit
in editor

Caeden117:
Forwarding comments/critiques from the ChroMapper discord:

> overall, there's just too many inconsistencies
for example, the margins between the refresh and the contributors and between the delete and save
example
why is one so close together and the other have some spacing?

> and the settings icons are inconsistent as well, some are filled, some are hollow, some are thin, some are thick, etc
example

> and for everything (I haven't tested the UI, only looked at pictures), anything interactable should react to user input
for example, when holding the modifier to modify how the mirror works (even though nothing happens in cm), the UI should change to represent the change, like changing the colors to be like my crappily drawn mirror icon
example
but that's just my two cents

>for me it's just the entire aesthetic, design and layout that all don't feel right
I also think the actual editor ui doesn't fit well with the rest at all
and the settings ui being so small makes it feel more like a mobile app than a desktop editor imo

My own opinions on this:

I think it's a good upgrade for the existing CM UI. I am not a UI person myself, I dont think I ever will be, so it's nice to have UI made by people who actually know how to do it. The theme system is also a nice touch for people to further customize CM's appearance, I'm excited to see how it could be used.

I am honestly fine with moving away from the older rainbow Chroma appearance for something that looks more... professional? mature? I've personally used exclusively Dark Theme ever since I implemented it, so the Chroma backgrounds haven't gotten much use for me. 

At the moment, the biggest problem I have with the proposal is actually in the song edit screen, where there's been some new additions made there that doesn't seem to be addressed in this PR. The big two that I immediately noticed is the lack of spectrogram preview, as well as the lack of file validation and an "open folder" button. I've attached screenshots of both of these in the current CM.

image

image

People also seem concerned over the lack of consistency if I were to accept this PR now, so I might wait for the other UI redesigns just to mass PR all at once.

Assets/__Scripts/UI/Themes/ThemeSO.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System;
+using TMPro;
+using UnityEngine;
+
+[CreateAssetMenu(fileName = "NewTheme", menuName = "ScriptableObjects/Theme", order = 1)]
+public class ThemeSO : ScriptableObject
+{
+    public Color backgroundPrimary;
```
Caeden117, line 8: Would it hurt much to switch these to `PascalCase` to fit .NET Naming Guidelines? Same question also applies to other classes in the PR.

Top-Cat:
Personally I think this PR is too big. I agree that the song list menu needs a facelift, it was next on my list and the editor UI has been on the to-do list for a while I know.

But you've combined that change with a new style that I'm personally not a fan of, especially how it applies to the song edit menu which in my opinion is a massive step backwards from the current design. I think it would have been better if you led with the theming features and applied them to an existing menu first. That way we can add that system while not making large visual changes and then separately PR any specific menu redesigns.

I could also point out number of code changes I'd want to see but for me this PR doesn't work in its current form.

monkeymanboy:
> At the moment, the biggest problem I have with the proposal is actually in the song edit screen, where there's been some new additions made there that doesn't seem to be addressed in this PR. The big two that I immediately noticed is the lack of spectrogram preview, as well as the lack of file validation and an "open folder" button. I've attached screenshots of both of these in the current CM.

The screenshots of the other menus aren't final and are a few months old so yeah they are lacking a few of the newer features, this would be fixed when I actually make the PR. I can see wanting to wait for the rest of the menus to accept the PR.
> I think it would have been better if you led with the theming features and applied them to an existing menu first

The main reason I didn't do the theming system separately is that it was built around the new UI and couldn't be applied to the old UI.
My main question now is should I even bother with making a PR for the rest?

Caeden117:
> The main reason I didn't do the theming system separately is that it was built around the new UI and couldn't be applied to the old UI.

Just quickly glancing at the changes, it *looks* like there should be little to no hassle in just slapping components on UI elements, so porting the theme system back to the previous CM UI *should* work.

> My main question now is should I even bother with making a PR for the rest?

I'll leave this up to you. I personally don't care, however it looks like top cat *might* care if you pull all of them into one mega PR

Top-Cat:
> My main question now is should I even bother with making a PR for the rest?

I certainly wouldn't make it one big PR (and one big commit at that, commit little and often). There are lots of things I like here but I also don't see the need to fundamentally redesign most of the UI.

I just think the idea of needing to redesign the UI to facilitate a code feature like theming is backwards.

A few people, including myself, did comment positively on the editor UI though so I'd certainly focus on that before any other menus.

monkeymanboy:
> I just think the idea of needing to redesign the UI to facilitate a code feature like theming is backwards.
The redesign wasn't made because of theming, I just meant that this color scheme with the 3 accent and 3 background colors doesn't apply to the current UI and so I didn't want to PR that first since it would need to be changed anyway.

> I certainly wouldn't make it one big PR (and one big commit at that, commit little and often)
Yeah, each menu would be its own PR I just mean in general. If the problem is with the design itself I don't wanna spend the time making the PR.

I'll go ahead and do the editor UI next though and just leave the rest alone for now.

PR #207, opened 2020-12-05
Fix errors during autosave clean

Top-Cat:
Autosave cleanup was always deleting the most recent autosave as even though we created an ordered list we never used it


Caeden117 (approved):
its a 10-line diff, what do you expect me to say, "k"?

PR #208, opened 2020-12-05
Schedule hit sounds instead of playing during frames

Top-Cat:
Basically moves the point at which are processed 0.5 seconds before they hit the play head and then adds a delay before they are played in the call to unity's audio system

Plus some handling for the notes between now and 0.5s time when pressing play and making sure things don't play after you stop


Top-Cat:
When trying to play 1/64ths:
image

image

Wouldn't call it perfect, but I don't know how to do any better than this

Caeden117 (approved):
I didn't hear any significant difference, but I'll take your word for it and hopefully there'll be people who will tell the difference

PR #210, opened 2020-12-06
in editor ui redesign

monkeymanboy:
Included is a complete redesign of the in-editor top UI.
Capture

Along with this, there are a few keybind changes to go along with this new UI.
This redesign removes the concept of a separate lighting workflow and so the tab button has been repurposed to pop out a side panel for less frequently used buttons (strobe gen and chroma rgb). A new bind has also been created for swapping between the two cursor intervals (default bind for this is X).


Assets/__Scripts/MapEditor/UI/Placement Controller UI/ColorTypeController.cs
```diff
@@ -0,0 +1,59 @@
...
+    [SerializeField] private Image rightLight;
+
+    void Start()
+    {
+        leftSelected.enabled = true;
+        rightSelected.enabled = false;
+        StartCoroutine(SetupColors());
+    }
+
+    private IEnumerator SetupColors()//PlatformDescriptor is in another scene so we have to find it, can probably change this get the colors from somewhere else in the future
```
Caeden117, line 24: the static `LoadInitialMap.PlatformLoadedEvent` does exist, so this enumerator can pretty much be replaced by a subscription to that event.

it's a bit weird, especially considering the class it resides in, but this will all be remedied with the zenject implementation

Assets/__Scripts/MapEditor/UI/Placement Controller UI/ColorTypeController.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        PlatformDescriptor descriptor = null;
+        while(descriptor == null)
+        {
+            descriptor = Resources.FindObjectsOfTypeAll<PlatformDescriptor>().FirstOrDefault(x => x.isActiveAndEnabled);
+            yield return new WaitForFixedUpdate();
+        }
+        leftNote.color = descriptor.colors.RedNoteColor;
+        leftLight.color = descriptor.colors.RedColor;
```
Caeden117, line 33: Another thing to note: Colors can be overridden by the user using the Chroma color panel, so this should probably be hooked up somehow.

*Another* another thing to note: Perhaps consider automatically switching the Light colors between the Boost/Non-Boost versions. I think that will be a nice touch and something people can appreciate.

Assets/Locales/Mapper_en.asset
```diff
@@ -675,6 +675,46 @@ MonoBehaviour:
...
+      m_Items: []
+  - m_Id: 147
+    m_Localized: Delete
+    m_Metadata:
+      m_Items: []
+  - m_Id: 148
+    m_Localized: Fade
+    m_Metadata:
+      m_Items: []
+  - m_Id: 149
```
Caeden117, line 698: Nitpick: Perhaps apply the English localization text to the other languages so people wouldn't be met with errors similar to `Missing localization for key "bar.events.on"`

Assets/__Scripts/MapEditor/UI/CurrentSectionDisplay.cs
```diff
@@ -0,0 +1,58 @@
...
+        if (isPlaying)
+        {
+            if (upcomingBookmarks.Count == 0)
+                return;
+            if (upcomingBookmarks.Peek().data._time <= atsc.CurrentBeat)
+                textMesh.text = upcomingBookmarks.Pop().data._name;
+        }
+        else
+        {
+            textMesh.text = bookmarkManger.bookmarkContainers.LastOrDefault(x => x.data._time <= atsc.CurrentBeat)?.data._name ?? "";
```
Caeden117, line 44: Very small performance nitpick: `.Where(predicate).LastOrDefault()` has been proven to be faster than `.LastOrDefault(predicate)`, but since this only happens when the user is not playing the song, I guess it's not *too* bad.

Caeden117 (commented):
Here is some feedback from just looking at the code. Full disclosure: I have not tested locally yet.

Top-Cat:
A couple of things to think about, none are show stoppers I'm happy (other than Caeden's comments) with this

- As there's no mode switching anymore the waveform seems to be locked on the left hand side
- Much like the refresh button a little more visual change on the mirror button. Someone made a cool concept in discord: Mirror
- Was the z-order change on the past notes grid intentional?

monkeymanboy:
> As there's no mode switching anymore the waveform seems to be locked on the left hand side
Could add a button for this on the new right panel (got any good ideas for an icon for said button?)
> Was the z-order change on the past notes grid intentional?
Yeah was just to get it rendering over the gradient background, does this cause problems somewhere I didn't notice?

Top-Cat:
> Yeah was just to get it rendering over the gradient background, does this cause problems somewhere I didn't notice?

The notes are now behind the grid

monkeymanboy:
That should cover everything, the swap spectrogram side icon could be better but it'll do for now.

PR #211, opened 2020-12-06
Hitsounds round 2

Top-Cat:
Fix bongo boi

Don't factor in time correction when scheduling sounds, I think this is what was causing the sounds to be slightly early as they were being scheduled as if deltaTime had passed?

Waveform is certainly more consistent

image


Caeden117 (approved):
looks good, would need some testers when this is built just to confirm

PR #215, opened 2020-12-12
Performance optimisations

Top-Cat:
Replace TMP in events with TM for performance. Event containers get created a lot faster as `CreateNewObject()` immediately sets the object as not active which causes the TMP to remove itself from a list in TMP_UpdateManager. As it's only just been created it will always be at the end of the list and do a full list traversal which adds up when there are a large number of objects visible.

Top is new. Only visible if you're looking for it.

image


Caeden117 (approved):
K

PR #216, opened 2020-12-16
Bugfixes

Top-Cat:
Fix error when packaging song with slashes in the name by removing slashes first
See: https://discord.com/channels/702230714585710602/702232127755780197/787542882357477426

Fix NPE when saving a difficulty
See: https://discord.com/channels/702230714585710602/702232127755780197/788448694986735656

Bonus tiny optimisation to event container Awake


Caeden117 (approved):
yeah should probably not ignore these while im working on other stuff

PR #219, opened 2021-01-04
Kda ground spikes

Top-Cat:
Not prop id but light id! Don't get excited, it only works for the spikes.

Bonus fix for being able to press keys in the strobe generator

image

### DIFF
--- a/Assets/__Scripts/MapEditor/Grid/Collections/EventsContainer.cs
+++ b/Assets/__Scripts/MapEditor/Grid/Collections/EventsContainer.cs
@@ -1,6 +1,7 @@
 ﻿using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using JetBrains.Annotations;
 using UnityEngine;
 using UnityEngine.InputSystem;
 
@@ -25,21 +26,34 @@ public class EventsContainer : BeatmapObjectContainerCollection, CMInput.IEventG
     public List<MapEvent> AllRotationEvents = new List<MapEvent>();
     public List<MapEvent> AllBoostEvents = new List<MapEvent>();
 
-    public bool PropagationEditing
+    public enum PropMode
+    {
+        Off, Prop, Light
+    }
+    
+    [CanBeNull]
+    public static string GetKeyForProp(PropMode mode) {
+        if (mode == PropMode.Light) return "_lightID";
+
+        return mode == PropMode.Prop ? "_propID" : null;
+    }
+    
+    public PropMode PropagationEditing
     {
         get { return propagationEditing; }
         set
         {
             propagationEditing = value;
             boxSelectionPlacementController.CancelPlacement();
-            int propagationLength = platformDescriptor.LightingManagers[EventTypeToPropagate]?.LightsGroupedByZ?.Length ?? 0;
-            labels.UpdateLabels(value, EventTypeToPropagate, value ? propagationLength + 1 : 16);
-            eventPlacement.SetGridSize(value ? propagationLength + 1 : SpecialEventTypeCount + platformDescriptor.LightingManagers.Count(s => s != null));
+            var propagationLength = (value == PropMode.Light ? platformDescriptor.LightingManagers[EventTypeToPropagate]?.ControllingLights?.Count :
+                platformDescriptor.LightingManagers[EventTypeToPropagate]?.LightsGroupedByZ?.Length) ?? 0;
+            labels.UpdateLabels(value, EventTypeToPropagate, value == PropMode.Off ? 16 : propagationLength + 1);
+            eventPlacement.SetGridSize(value != PropMode.Off ? propagationLength + 1 : SpecialEventTypeCount + platformDescriptor.LightingManagers.Count(s => s != null));
             EventTypePropagationSize = propagationLength;
             UpdatePropagationMode();
         }
     }
-    private bool propagationEditing = false;
+    private PropMode propagationEditing = PropMode.Off;
 
     private void Start()
     {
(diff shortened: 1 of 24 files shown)


Assets/__Scripts/MapEditor/Grid/PauseToggleLights.cs
```diff
@@ -12,8 +13,15 @@ public class PauseToggleLights : MonoBehaviour
     private MapEvent defaultBoostEvent = new MapEvent(0, 5, 0);
 
     private const int NOT_PROP = -1;
-    private Dictionary<int, Dictionary<int, MapEvent>> lastEvents = new Dictionary<int, Dictionary<int, MapEvent>>();
+    private Dictionary<int, LastEvents> lastEvents = new Dictionary<int, LastEvents>();
     private List<MapEvent> lastChromaEvents = new List<MapEvent>();
+    
+    private class LastEvents
+    {
+        [CanBeNull] public MapEvent lastEvent = null;
```
Caeden117, line 21: See other comment regarding the `[CanBeNull]` attribute

Assets/__Scripts/MapEditor/UI/Strobe Generator/StrobeTextField.cs
```diff
@@ -0,0 +1,8 @@
+﻿using TMPro;
+using UnityEngine;
+
+[RequireComponent(typeof(TMP_InputField))]
+public class StrobeTextField : DisableActionsField
```
Caeden117, line 5: Honestly, this seems like something useful enough to where renaming into something like `DisableActionsOnInteract` would make sense.

Assets/__Scripts/Platforms/PlatformDescriptor.cs
```diff
@@ -260,6 +260,21 @@ void HandleLights(LightsManager group, int value, MapEvent e)
         if (SoloAnEventType && e._type != SoloEventType) mainColor = invertedColor = Color.black.WithAlpha(0);
 
         IEnumerable<LightingEvent> allLights = group.ControllingLights;
+
+        if ((e._customData?.HasKey("_lightID") ?? false) && Settings.Instance.EmulateChromaAdvanced)
+        {
+            int lightID = group.EditorToGameLightIDMap.IndexOf(e._customData["_lightID"].AsInt);
+            if (lightID >= 0 && lightID < group.ControllingLights.Count)
+            {
+                allLights = new List<LightingEvent> {group.ControllingLights[lightID]};
```
Caeden117, line 269: Is constructing a new list with every lightID event necessary? From experience doing something similar to this in previous ChroMapper versions, `List<T>`s can be pretty expensive to construct. We should be able to just assign straight into the `allLights` variable because it is a `IEnumerable<LightingEvent>`.

Assets/__Scripts/Platforms/PlatformDescriptor.cs
```diff
@@ -260,6 +260,21 @@ void HandleLights(LightsManager group, int value, MapEvent e)
...
+            }
+            else
+            {
+                Debug.LogWarning($"Light ID {lightID} does not exist for event type {e._type}!");
+                allLights = new List<LightingEvent>();
```
Caeden117, line 274: I do realize that this exact same thing is being done a little bit down the page, but we can make a smaller allocation by using `Enumerable.Empty<LightingEvent>()`

Assets/__Scripts/Platforms/PlatformDescriptor.cs
```diff
@@ -270,7 +285,7 @@ void HandleLights(LightsManager group, int value, MapEvent e)
             else
             {
                 Debug.LogWarning($"Light Prop ID {propID} does not exist for event type {e._type}!");
-                allLights = new List<LightingEvent>() { };
+                allLights = new List<LightingEvent>();
```
Caeden117, line 288: See other comment talking about `Enumerable.Empty<LightingEvent>()`

Caeden117 (commented):
This is just my own thoughts before locally testing these changes.

Caeden117 (changes requested):
I'm getting some uncaught exceptions when attempting to move/place events with light IDs above the 70-ish range with KDA's spike lights

image

Assets/__Scripts/MapEditor/UI/Strobe Generator/StrobeTextField.cs
Top-Cat: The point is that the DisableActionsField class calls `DisableActionMaps` based on a call to `GetType()`, so it's a separate class on purpose to isolate these from other things that call that method

Assets/__Scripts/MapEditor/Grid/Collections/EventsContainer.cs
```diff
@@ -25,21 +26,34 @@ public class EventsContainer : BeatmapObjectContainerCollection, CMInput.IEventG
     public List<MapEvent> AllRotationEvents = new List<MapEvent>();
     public List<MapEvent> AllBoostEvents = new List<MapEvent>();
 
-    public bool PropagationEditing
+    public enum PropMode
+    {
+        Off, Prop, Light
+    }
+    
+    [CanBeNull]
```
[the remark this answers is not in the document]
Top-Cat: By default it literally won't let me type `string?`. Afaik it does improve ReSharper's static analysis results, but at what cost.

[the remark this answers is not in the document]
Top-Cat: Oh... that's why
image

Caeden117 (approved):
Alright, thanks for fixing this for me 👍

PR #221, opened 2021-01-06
Add NE note scaling

Futuremappermydud:


Caeden117 (approved):
For future reference (pun not intended), I would like a newline of space between new methods, but enforcing coding guidelines will come later. Good job.

Futuremappermydud:
> For future reference (pun not intended), I would like a newline of space between new methods, but enforcing coding guidelines will come later. Good job.

Sorry, probably could’ve done a better job at cleaning it up.

PR #223, opened 2021-01-06
Bugfixes

Top-Cat:
Ignore null when copying map custom data.
Fixes: https://discord.com/channels/702230714585710602/702232127755780197/795505064641953792

Fix escape returning to main menu from mapper view

Revert change to alt+scroll that skips laser speed 4, works for me without this

Remove bind helper when leaving scene


Caeden117 (approved):
Alright, back to the usual: k

PR #224, opened 2021-01-08
Light id changes / fixes

Top-Cat:
I missed a spot in gradient gen
Add light id offset because KDA sticks can't be re-ordered sensibly in prop id mode due to groupings between the chevron / lane lights and random sticks. But we can adjust the light id order.


Caeden117:
test comment to make sure that discord webhook isn't being fucky

Caeden117 (approved):
congratulations, the PR was so dummy thicc that it broke the webhook in the CM discord

PR #225, opened 2021-01-09
Various features

Top-Cat:
I can break this out if needed.

Added KDA tentacles to objects that get hidden with 'L'
Allow bookmark renaming with right click
Add mirroring hotkeys:
- M = Same as UI button
- Shift+M = Colours only
- Alt+M = Mirror in time

Mirroring works on Chroma 2 gradients, swaps startColor and endColor

Changed it so that Chroma placment option is persistent and works even with the chroma menu closed


Caeden117 (approved):
k

PR #227, opened 2021-01-13
Add chroma step gradient easings including two custom easings

Top-Cat:
Made it way easier to scroll in the strobe generator UI by forwarding events down, adding an invisible background and increasing scroll speed.

Easings include all the normal tweens plus a flicker (first / last / first / last) and random (examples below)

Fixed a bug with nullable prop id not being checked and used prop mode instead
Should include fix for UI mirror button not working

image
image


Caeden117 (approved):
this also looks good

PR #228, opened 2021-01-13
Fix selection in 360

Top-Cat:
Transform bounds correctly when rotated into the co-ordinate space we use for selection, this was preventing (or sometimes allowing) the selection box getting to objects in the map


Caeden117 (approved):
k, looks good

PR #229, opened 2021-01-17
Keybind rejig + assorted other changes

Top-Cat:
- Moved/Added the following:
  - Precision placement toggle `Q`
  - Reset rings `Numpad 0`
  - Camera attach to note grid `Ctrl+R`
  - Leave options menu `Escape`
- Fixed rocket center prop id order
- Add chroma suggestion for environment removal, could bump this to a requirement?
- Fixed offset precision placement
- Removed waveform follows workflow option

### DIFF
--- a/Assets/_Prefabs/MapEditor/Platforms/Rocket.prefab
+++ b/Assets/_Prefabs/MapEditor/Platforms/Rocket.prefab
@@ -99,6 +99,7 @@ MonoBehaviour:
   CanBeTurnedOff: 1
   currentAlpha: 0
   multiplyAlpha: 1
+  lightIdOffset: 0
 --- !u!1 &345466861614610517
 GameObject:
   m_ObjectHideFlags: 0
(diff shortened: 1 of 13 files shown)


Caeden117 (changes requested):
Keybind proposals conflict a little bit here:
- `Ctrl+R` for camera attaching to note grid conflicts with `Ctrl+R` for precision 360/90 rotation
  - Not sure who cares about precision 360/90 rotation now, but it does still exist
- `Numpad 0` for resetting rings conflicts with `Alt + Numpad 0-9` for laser speed input
  - EDIT: While laser speed input does reply on `Alt` being held, due to the unique nature of how it's implemented, it will not block the Reset Rings action from firing

Top-Cat:
Precision 360/90 is just R not Ctrl+R?

Top-Cat:
And now I see that numpad conflicts for note diagonals too, hmm

Caeden117:
I dont think anyone would miss the dedicated diagonal note keys

(and yes Precision 360/90 is `R`, my bad)

Top-Cat:
I've added some workarounds that stop the numpad hotkeys working while laser speed editing is happening. For some reason this didn't work via disable / clearing action maps 🤷

Caeden117 (approved):
alright, looks good to me now

PR #230, opened 2021-01-20
Add experimental dsp buffer size option to allow reducing audio latency

Top-Cat:
Requires editor restart after changing.

Down to 256 seems to work fine for me, but having it as an options allows people with different hardware to try different options out. Don't know if this should remain an option of if we can just pick a lower number going forward.

### DIFF
--- a/Assets/__Scripts/UI/PersistentUI.cs
+++ b/Assets/__Scripts/UI/PersistentUI.cs
@@ -93,8 +93,18 @@ public class PersistentUI : MonoBehaviour {
         CMInputCallbackInstaller.PersistentObject(transform);
         LocalizationSettings.SelectedLocale = Locale.CreateLocale(Settings.Instance.Language);
         AudioListener.volume = Settings.Instance.Volume;
+
+        UpdateDSPBufferSize();
+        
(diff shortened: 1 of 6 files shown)


Assets/__Scripts/UI/PersistentUI.cs
```diff
@@ -93,10 +93,20 @@ private void Awake()
...
+        UpdateDSPBufferSize();
+        
         centerDisplay.host = this;
         bottomDisplay.host = this;
     }
 
+    private static void UpdateDSPBufferSize()
+    {
+        var config = AudioSettings.GetConfiguration();
+        config.dspBufferSize = (int) Math.Pow(2, Convert.ToInt32(Settings.Instance.DSPBufferSize));
```
Caeden117, line 106: Is there any reason why you are using `Convert.ToInt32()` when the variable you're trying to convert is... already an integer?

Top-Cat: You saw nothing

Caeden117 (approved):
alright, seems fine now

(NOTE: beat saber just updated so some time is gonna be spent updating counters+)

PR #231, opened 2021-01-23
Add song audio control

Top-Cat:
Add new slider to control song volume so that you can make metronome or hitsounds louder than the song.
Fix master volume not being set correctly due to audio reset.
Update translations.

### DIFF
--- a/Assets/Locales/FirstBoot_de.asset
+++ b/Assets/Locales/FirstBoot_de.asset
@@ -19,7 +19,7 @@ MonoBehaviour:
     m_Items: []
   m_TableData:
   - m_Id: 1
-    m_Localized: First Time Setup
+    m_Localized: Ersteinrichtung
     m_Metadata:
       m_Items: []
   - m_Id: 2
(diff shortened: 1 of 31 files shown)


Caeden117 (changes requested):
Changing Song audio should apply at runtime while in the editor

Top-Cat:
> Changing Song audio should apply at runtime while in the editor (EDIT: scene, not the Unity Editor)

It should already?

Caeden117:
I understand that it updates when you first play the song, but it should probably also update *while* the song is playing

Top-Cat:
> I understand that it updates when you ~first play~ start playing the song, but it should probably also update _while_ the song is playing

I figured people wouldn't be changing options while their song was playing and it made it 10x easier to implement. But if you want.

Caeden117:
If not, I *guarantee* you that someone would report it as a bug. It's an odd inconsistency that the other audio settings (Master, hit sounds, metronome) update as soon as the setting is changed, but the song audio setting doesn't. Not to mention people would expect the song audio to update as soon as the setting is changed

Top-Cat:
> If not, I _guarantee_ you that someone would report it as a bug. It's an odd inconsistency that the other audio settings (Master, hit sounds, metronome) update as soon as the setting is changed, but the song audio setting doesn't. Not to mention people would expect the song audio to update as soon as the setting is changed

Metronome actually doesn't work if starts disabled

Caeden117:
that should probably be fixed too then, but probably not this PR

PR #232, opened 2021-01-23
Only use _preciseSpeed when necessary

Pixelguymm:
Will not use _preciseSpeed if decimalPreciseSpeed is a whole number. Will also change _value to the whole number closest to decimalPreciseSpeed (unless that number is 0, in which case 1 will get used).

I changed one line and added another, probably badly, you're welcome.

### DIFF
--- a/Assets/__Scripts/MapEditor/Mapping/Strobe Generator/Passes/StrobeLaserSpeedInterpolationPass.cs
+++ b/Assets/__Scripts/MapEditor/Mapping/Strobe Generator/Passes/StrobeLaserSpeedInterpolationPass.cs
@@ -86,7 +86,8 @@ public class StrobeLaserSpeedInterpolationPass : StrobeGeneratorPass
             var decimalPreciseSpeed = Math.Round(Mathf.Lerp(lastSpeed, nextSpeed, easingFunc(progress)), decimalPrecision);
             MapEvent data = new MapEvent(newTime, type, 1);
             data._customData = new JSONObject();
-            data._customData["_preciseSpeed"] = decimalPreciseSpeed;
+            data._value = (int)(Math.Round(decimalPreciseSpeed, MidpointRounding.AwayFromZero) == 0 && decimalPreciseSpeed != 0 ? 1 : Math.Round(decimalPreciseSpeed, MidpointRounding.AwayFromZero));
+            if (decimalPreciseSpeed % 1 != 0) data._customData["_preciseSpeed"] = decimalPreciseSpeed;
             if (overrideDirection)
             {
                 switch (type)


Assets/__Scripts/MapEditor/Mapping/Strobe Generator/Passes/StrobeLaserSpeedInterpolationPass.cs
```diff
@@ -86,7 +86,8 @@ public override IEnumerable<MapEvent> StrobePassForLane(IEnumerable<MapEvent> or
             var decimalPreciseSpeed = Math.Round(Mathf.Lerp(lastSpeed, nextSpeed, easingFunc(progress)), decimalPrecision);
             MapEvent data = new MapEvent(newTime, type, 1);
             data._customData = new JSONObject();
-            data._customData["_preciseSpeed"] = decimalPreciseSpeed;
+            data._value = (int)(Math.Round(decimalPreciseSpeed, MidpointRounding.AwayFromZero) == 0 && decimalPreciseSpeed != 0 ? 1 : Math.Round(decimalPreciseSpeed, MidpointRounding.AwayFromZero));
```
Caeden117, line 89: This can easily be simplified to something like

image

Top-Cat:
Is this supposed to handle negative laser speeds? The use of away from zero seems a little weird to me.

Caeden117:
> 
> 
> Is this supposed to handle negative laser speeds? The use of away from zero seems a little weird to me.

No, but it corrects some potentially unwanted behavior.

This table was taken from the [.NET documentation for the Round method](https://docs.microsoft.com/en-us/dotnet/api/system.midpointrounding?view=net-5.0):

image

Top-Cat:
I'm not saying the default is necessarily correct either, just questioning if AwayFromZero is. If it's supposed to handle negative values I'd think -0.5 shouldn't result in a speed of 1.

Assets/__Scripts/MapEditor/Mapping/Strobe Generator/Passes/StrobeLaserSpeedInterpolationPass.cs
Pixelguymm: That will still change a decimalPreciseSpeed of 0 to 1, right? Other than that it's definitely better.

Caeden117: Yes, the `Max` method would take the higher value of the two numbers. With a decimalPreciseSpeed of 0 (and below), it will return 1

Top-Cat: This solution doesn't handle negative numbers (see discussion below)

Caeden117: tbf, i dont think we need to support negative numbers, since they'd behave the exact same as positive numbers (with the direction flipped)

Top-Cat: Well that's the exact question I asked.

Pixelguymm: 0 should probably stay as 0, not default to 1.
And like Caeden said, I didn't think handling negative numbers would be necessary.

Caeden117:
Is this PR still something you want merged?

Pixelguymm:
It's still a feature that I think should be implemented, but if you want me to implement it some other way (or want to do it yourself), you can close the PR.

PR #233, opened 2021-01-24
Save notification doesn't fade, disappears early when save is complete

Top-Cat:
Options available for other notifications, I've extended the wait time for save as the fade would previously have been 2 seconds.

### DIFF
--- a/Assets/__Scripts/UI/PersistentUI.cs
+++ b/Assets/__Scripts/UI/PersistentUI.cs
@@ -117,30 +117,35 @@ public class PersistentUI : MonoBehaviour {
         Settings.Instance.Save();
     }
 
-    public void DisplayMessage(string message, DisplayMessageType type) {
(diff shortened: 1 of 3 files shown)


Assets/__Scripts/UI/PersistentUI.cs
```diff
@@ -386,42 +391,90 @@ public enum DialogBoxPresetType
...
 
+        public class NotificationMessage
+        {
+            private AsyncOperationHandle<string>? _localisable;
+
+            public float waitTime = 2.0f;
+            public string message;
+            public bool cancelled = false;
+            public bool skipFade = false;
+            public readonly DisplayMessageType type;
```
Caeden117, line 402: This isn't going to be a PR killer, but I will say that sometime after Zenject, I will be more strict with the [.NET Naming Guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/naming-guidelines)

(TL;DR: please use `PascalCase` with public members, kthx)

PR #236, opened 2021-02-06
Objects actions refactor

Top-Cat:
I didn't really want to make any huge changes during zenjectification but this kinda happened.
Makes most undo/redo actions faster by not deleting/spawning objects when we don't need to.
Huge boost to general delete performance by reverting back to before we tried to add a load of patches to fix ghost notes.
Pool refreshes on multi-actions are saved until the end of all actions.
21 new tests (350% increase) to validate most ways of editing objects do the right thing and can be undone and redone (and don't create ghost objects).
A lot of fixes to object editing found through testing.
Colours now always contain an alpha value, this makes node editor consistently show them when selecting multiple objects.

Bonus fix for playing after a gradient causing flashbangs.

### DIFF
--- a/Assets/Tests/BPMTest.cs
+++ b/Assets/Tests/BPMTest.cs
@@ -1,0 +1,65 @@
+﻿using NUnit.Framework;
+using System.Collections;
+using System.Linq;
+using Tests.Util;
+using UnityEngine;
+using UnityEngine.TestTools;
+
+namespace Tests
+{
+    public class BPMTest
+    {
+        [UnityOneTimeSetUp]
+        public IEnumerator LoadMap() => TestUtils.LoadMapper();
+
+        [TearDown]
+        public void ContainerCleanup()
+        {
+            BeatmapActionContainer.RemoveAllActionsOfType<BeatmapAction>();
+            TestUtils.CleanupBPMChanges();
+        }
+
+        private static void CheckBPM(BeatmapObjectContainerCollection container, int idx, int time, int bpm)
+        {
+            var newObjA = container.LoadedObjects.Skip(idx).First();
+            Assert.IsInstanceOf<BeatmapBPMChange>(newObjA);
+            if (newObjA is BeatmapBPMChange newNoteA)
+            {
+                Assert.AreEqual(time, newNoteA._time);
+                Assert.AreEqual(bpm, newNoteA._BPM);
+            }
+        }
+        
+        [Test]
+        public void ModifyEvent()
+        {
+            var actionContainer = Object.FindObjectOfType<BeatmapActionContainer>();
+            var collection = BeatmapObjectContainerCollection.GetCollectionForType(BeatmapObject.Type.BPM_CHANGE);
+            if (collection is BPMChangesContainer bpmCollection)
+            {
+                var root = bpmCollection.transform.root;
+
+                var bpmChange = new BeatmapBPMChange(50, 10);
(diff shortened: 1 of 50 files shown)


Assets/Tests/BPMTest.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+        
+        [Test]
+        public void ModifyEvent()
+        {
+            var actionContainer = Object.FindObjectOfType<BeatmapActionContainer>();
+            var collection = BeatmapObjectContainerCollection.GetCollectionForType(BeatmapObject.Type.BPM_CHANGE);
+            if (collection is BPMChangesContainer bpmCollection)
+            {
+                var root = bpmCollection.transform.root;
```
Caeden117, line 40: i am root

(is there a reason why we're getting the root here? i'm not seeing it used anywhere)

Assets/__Scripts/Map/BeatmapObject.cs
```diff
@@ -100,4 +102,5 @@ public virtual bool IsConflictingWith(BeatmapObject other, bool deletion = false
         }
         return false;
     }*/
+    public abstract void Apply(BeatmapObject originalData);
```
Caeden117, line 105: I think some code could be re-used here if you made this method `virtual`. We could have the `_time` and `_customData` applications be done here, and the overriding methods could just call `base.Apply(originalData);` to re-use this code. Just personal preference, though

Assets/__Scripts/MapEditor/Grid/Collections/BeatmapObjectContainerCollection.cs
```diff
@@ -237,7 +237,9 @@ public void RemoveConflictingObjects(IEnumerable<BeatmapObject> newObjects, out
         foreach (BeatmapObject newObject in newObjects)
         {
             Debug.Log($"Performing conflicting check at {newObject._time}");
-            BeatmapObject conflict = UnsortedObjects.Find(x => x.IsConflictingWith(newObject));
+
+            var localWindow = GetBetween(newObject._time - 0.1f, newObject._time + 0.1f);
+            BeatmapObject conflict = localWindow.FirstOrDefault(x => x.IsConflictingWith(newObject));
```
Caeden117, line 242: last i've checked, a straight `FirstOrDefault(predicate)` check is slower than a `.Where(predicate).FirstOrDefault()`. Does that apply here? I'm pretty sure the former does a bit extra processing to ensure that the match it found really is the first in the collection.

Caeden117 (commented):
Just some general feedback so far, before I locally pull the changes and test them

Assets/__Scripts/MapEditor/Grid/Collections/BeatmapObjectContainerCollection.cs
Top-Cat: localWindow should be small enough for this to not really matter but I've changed it anyway

Assets/__Scripts/Map/BeatmapObject.cs
Top-Cat: Done

Assets/Tests/BPMTest.cs
Top-Cat: Copy paste fail

Caeden117 (approved):
alright, been testing on and off for the past few days, this appears to be all good

PR #237, opened 2021-02-12
Add logging around failing test

Top-Cat:
WaitForEndOfFrame hangs forever in batch tests, null gets run slightly earlier in the loop but it seems to load the lights fine.

Any thoughts if this could have any other side effects?


Caeden117 (approved):
k

PR #238, opened 2021-02-13
Bug fixes

Top-Cat:
Fixed being able to interact with pyramid collider as we were only removing box colliders. Removed DestroyBoxCollider as we no longer need the collider on the selection box.

Update reset to match current chroma (looks to have been changed 6 months ago)


Caeden117 (approved):
aight

PR #242, opened 2021-02-17
Only apply big offset only if validation is likely

Top-Cat:


Caeden117 (approved):
this seems simple enough, surely nothing will ever break from this

PR #244, opened 2021-02-27
My attempt at fixing audio sync late into songs

Top-Cat:
I'm not super confident with what's going on in this method but it seems like using the time ratios will approach 1 as the current time tends towards infinity. This means for large time values their distance from 1 will never be >= smoothDeltaTime. This is bad as it means we can de-sync and never correct for it. I _think_ most of this desync in the map I'm testing is caused as playback starts (putting events into the queue, finding the current lighting state etc.) so if we can trigger a single correction this is all we need.

I think this may also be related to the skipping issue some people have reported and while I can't really reproduce it I hope this may improve that behaviour.

I've also replaced the usage of songAudioSource.time with timeSamples, this is marginally more accurate and is probably unnecessary but I'd already been testing with it anyway.


Caeden117 (approved):
I'm not the kind of person who notices small audio delays, so I'll take your word for it here

(hopefully it works for anyone who does actually notice them)

PR #245, opened 2021-02-28
Bug fixes

Top-Cat:
- CM-61 - Improved the check in the new level dialog and locked the folder name to match so we don't have TOCTOU problems
- Split undo / redo bindings into separate actions as otherwise the InputSystemPatch doesn't work properly (Ctrl + Shift + Z was undoing and then redoing for me while testing bpm changes)
- Fix BPM changes not updating the grid when undoing/redoing/pasting
- Fix an error from the folder browser dialog I got on OSX relating to the playerloop. Internet says the async methods are better 🤷 and fix misalignment of the text on the first boot screen (I really hate this unity bug)


Caeden117 (approved):
aight, looks good to me chief

PR #246, opened 2021-03-06
Bug fixes

Top-Cat:
CM-66 - Difficulty renaming bug
Fix undo button in song info ui not working (it was tied to contributors only?)
Fix moving after using playing view and pressing movement keys while in playing view


Top-Cat:
Well that was a fun adventure, accidentally lost and then refound those changes. Plus fixed pressing tab after opening options menu.

Caeden117 (approved):
LGTM

PR #247, opened 2021-03-10
1.13.4 env

Top-Cat:
Prop reordering for new update

Still need to test some of these as KDA took way too long and I rushed through the rest


Caeden117 (approved):
LGTM, cant wait for these to be broken next update 😃

Top-Cat:
Ready for some testing

Top-Cat:
Now with kone

Caeden117 (approved):
kalcbwujavgcuscope

PR #248, opened 2021-03-12
Include wav support

Fernthedev:
This just makes it so `.wav` actually gets loaded

### DIFF
--- a/Assets/__Scripts/UI/SongInfoEditUI.cs
+++ b/Assets/__Scripts/UI/SongInfoEditUI.cs
@@ -32,6 +32,13 @@ public class SongInfoEditUI : MenuBase
         }
     }
 
+    private static Dictionary<string, AudioType> ExtensionToAudio = new Dictionary<string, AudioType>()
+    {
+        {"ogg", AudioType.OGGVORBIS},
+        {"egg", AudioType.OGGVORBIS},
+        {"wav", AudioType.WAV}
+    };
+
     public static List<Environment> VanillaEnvironments = new List<Environment>()
     {
         new Environment("Default", "DefaultEnvironment"),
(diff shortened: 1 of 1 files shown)


Fernthedev:
`.egg` files are not working, will be fixed in a later commit

Assets/__Scripts/UI/SongInfoEditUI.cs
```diff
@@ -285,10 +292,15 @@ private IEnumerator LoadAudio(bool useTemp = true)
...
-            if (audioPath.text.ToLower().EndsWith("ogg") || audioPath.text.ToLower().EndsWith("egg"))
+
+            var extension = audioPath.text.Contains(".") ? Path.GetExtension(audioPath.text.ToLower()).Replace(".", "") : "";
+
+
+            if (!string.IsNullOrEmpty(extension))
             {
                 Debug.Log("Lets go");
-                UnityWebRequest www = UnityWebRequestMultimedia.GetAudioClip($"file:///{Uri.EscapeDataString($"{fullPath}")}", AudioType.OGGVORBIS);
+                var audioType = ExtensionToAudio.ContainsKey(extension) ? ExtensionToAudio[extension] : AudioType.UNKNOWN;
```
Caeden117, line 302: Would it be a good idea to allow users to continue into the editor if the audio type is unknown? While Unity *might* support a few other audio formats, BeatSaver only supports `.ogg`, `.egg`, and `.wav` (I checked the source code). Furthermore, unsupported audio formats will show up as no audio at all, and I can guarantee you that it would be reported as a bug sometime in the future.

Fernthedev: I was unsure of this and assumed that might be a possibility. Thanks for confirming, will be taken into account

Caeden117 (approved):
LGTM

PR #250, opened 2021-03-21
Rocket league lightid

Top-Cat:
Rocket league has one in-game light id that maps to a number of lights in CM. A special environment indeed.
Could try and make a new mesh for the light but this was easier.


Caeden117 (approved):
image

PR #254, opened 2021-03-27
Added rotate in place keybinds

monkeymanboy:
Adds three new keybinds, one for left rotations one for right (E and Q by default), as well as a modifier (Shift by default). These keybinds are for inserting/replacing rotation events at the current location in the song. The modifier keybind is used to switch between late and early rotation events. Currently only supports base game rotation events in increments of 15 degrees. Feel free to change the default keybinds, E/Q just made the most sense but not sure if there are plans to put something else on those buttons.

### DIFF
--- a/Assets/__Scripts/Input/Master.cs
+++ b/Assets/__Scripts/Input/Master.cs
@@ -1007,6 +1007,30 @@ public class @CMInput : IInputActionCollection, IDisposable
                     ""expectedControlType"": """",
                     ""processors"": """",
                     ""interactions"": ""Press""
+                },
+                {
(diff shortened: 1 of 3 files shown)


Top-Cat:
I think Q is currently bound to precision placement by default. I'm not against moving this to something else (in fact I think it's probably a good idea) but probably best to not just break it :)

Caeden117 (commented):
This looks good so far, however I agree with TopCat; let's move the current Precision Placement keybind somewhere else to make room for rotate in place. I'm unhappy with the current implementation so I wouldn't really mind where you put it.

PR #255, opened 2021-03-28
Holding Play and releasing after X time rewinds time

TheRandomosity:
Bringing in behavior from MMA allowing for releasing the play button to rewind back to where playing began IF play was held for a given length of time.

I've been using it for a couple days no problem. Not sure if there's any dependencies I might have missed though, or if I am not following any coding standard properly, etc. Lemme know!

### DIFF
--- a/Assets/__Scripts/MapEditor/AudioTimeSyncController.cs
+++ b/Assets/__Scripts/MapEditor/AudioTimeSyncController.cs
@@ -61,6 +61,7 @@ public class AudioTimeSyncController : MonoBehaviour, CMInput.IPlaybackActions, 
 
     public bool IsPlaying { get; private set; }
     private float playStartTime;
+    private const float cancelPlayInputDuration = 0.3f;
 
     private float offsetMS;
     public float offsetBeat { get; private set; } = -1;
@@ -261,8 +262,7 @@ public class AudioTimeSyncController : MonoBehaviour, CMInput.IPlaybackActions, 
         if (context.performed) TogglePlaying();
 
         // if play is held and released a significant time later, cancel playing instead of merely toggling
-        const float cancelInputDuration = 0.3f;
-        if (context.canceled && context.duration >= cancelInputDuration) CancelPlaying();
+        if (context.canceled && context.duration >= cancelPlayInputDuration) CancelPlaying();
     }
 
     public void OnResetTime(InputAction.CallbackContext context)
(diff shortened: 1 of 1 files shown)


Assets/__Scripts/MapEditor/AudioTimeSyncController.cs
```diff
@@ -247,6 +259,10 @@ public void RefreshGridSnapping()
     public void OnTogglePlaying(InputAction.CallbackContext context)
     {
         if (context.performed) TogglePlaying();
+
+        // if play is held and released a significant time later, cancel playing instead of merely toggling
+        const float cancelInputDuration = 0.3f;
```
Caeden117, line 264: is there a reason why this constant is defined in the method? wouldn't hurt to move it out of the method

TheRandomosity: Just personal preference, moved it up next to `playStartTime`.

Assets/__Scripts/MapEditor/AudioTimeSyncController.cs
```diff
@@ -192,6 +195,16 @@ private void OnDestroy()
...
+    public void CancelPlaying()
+    {
+        IsPlaying = false;
+        CurrentSeconds = playStartTime;
+
+        songAudioSource.Stop();
+        SnapToGrid();
+
+        if (OnPlayToggle != null) OnPlayToggle(IsPlaying);
+    }
```
Caeden117, line 207: I feel like this is re-inventing the wheel a little bit. Surely this method can't be replaced with a `TogglePlaying` call, then followed by setting `CurrentSeconds`? Maybe add an `IsPlaying` check just to be extra sure.

TheRandomosity: Sorry it took so long to get back to ya, it's been a busy week.
I vaguely remember doing it this way in an effort to maintain the same order of operations. But present me isn't sure what past me was thinking.
Would it make more sense to reset CurrentSeconds _before_ calling TogglePlaying though? I've tested both ways, either works, but to prevent any future changes to SnapToGrid or anything listening to OnPlayToggle that might check CurrentSeconds from breaking.

Top-Cat: Well, what happens if I'm in the middle of a beat but set my precision to 1/1 and then press, hold and release?
Surely depending on the order here I'll either return to the middle of the beat or return to a snapped beat?

TheRandomosity: Ahhh, SnapToGrid, gotcha. You are correct. I would say returning to exactly where I was would be the preferred option. For the same reason that if I change my precision, I don't immediately get snapped to the grid - it's unexpected behavior if anything other than myself changes the position.

PR #257, opened 2021-04-04
Add quick note editing option

Top-Cat:


Assets/__Scripts/MapEditor/Input/BeatmapNoteInputController.cs
```diff
@@ -34,6 +34,7 @@ public class BeatmapNoteInputController : BeatmapInputController<BeatmapNoteCont
     };
 
     [SerializeField] private NoteAppearanceSO noteAppearanceSO;
+    public static bool QuickModificationActive;
```
Caeden117, line 37: does this need to be static? It would persist when the scene is closed; if you hold alt while exiting, and edit a new song, CM will think you are still quick editing until you press alt again

Assets/__Scripts/MapEditor/AudioTimeSyncController.cs
```diff
@@ -139,7 +139,8 @@ private void OnDestroy()
                 var correction = CurrentSeconds > 1 ? trackTime / CurrentSeconds : 1f;
 
                 // Snap forward if we are more than a 2 frames out of sync as we're trying to make it one frame out?
-                if (Mathf.Abs(trackTime - CurrentSeconds) >= 2 * Time.smoothDeltaTime * (songSpeed / 10f))
+                float frameTime = Mathf.Max(0.04f, Time.smoothDeltaTime * 2);
+                if (Mathf.Abs(trackTime - CurrentSeconds) >= frameTime * (songSpeed / 10f))
```
Caeden117, line 143: These changes appear to be exactly identical to #259. Is this intentional?

Top-Cat: Nope, branching fail

PR #258, opened 2021-04-04
Persist cursor precision

Top-Cat:


Assets/__Scripts/MapEditor/UI/PrecisionStepDisplayController.cs
```diff
@@ -12,17 +12,26 @@ public class PrecisionStepDisplayController : DisableActionsField
     [SerializeField] private Color defaultOutlineColor;
     [SerializeField] private Color selectedOutlineColor;
 
-    private bool firstActive;
+    private bool _firstActive;
```
Caeden117, line 15: out of curiosity, why was there an underscore added to this field?

Top-Cat: https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md #3 🤷‍♂️
I can change it back not really a big deal

romandamen: nah underscore is better

PR #262, opened 2021-04-11
BPM Fixes

Top-Cat:
Seems node editor doesn't update the grid if you use it to change bpm changes.
But the code for that is already in actions so just switch to using that code.

Also confused why `_customData: null` was being included in node editor so added a null check to ConvertToJSON in BeatmapBPMChange


Caeden117 (approved):
hey at least this makes the node editor code a bit nicer

also yet another banger

PR #263, opened 2021-04-13
More accurate audio preview

XAce1337manX:
Currently, this is how the preview sounds like:
image

Now, it looks like this:
image


Top-Cat:
It may not be accurate but I remember not fading in to be quite jarring.

XAce1337manX:
I personally prefer the accuracy as it's what it sounds like in game. 
Removing the fade in is also easily revertible if others also find it jarring.

XAce1337manX:
Turns out the fade in depends on if you have the background menu music on or not. It's also inconsistent. Out of the times it did fade in, it was a tad faster than before.

PR #267, opened 2021-04-18
Extension buttons

monkeymanboy:
This PR adds an easy way for plugins to add buttons to the right side pop out panel of buttons in the editor. Usage is described in PLUGINS.md.

### DIFF
--- a/Assets/__Scripts/MapEditor/UI/Extensions/ExtensionButtons.cs
+++ b/Assets/__Scripts/MapEditor/UI/Extensions/ExtensionButtons.cs
@@ -13,6 +13,11 @@ public static class ExtensionButtons
         extensionButton.Tooltip = tooltip;
         extensionButton.Icon = icon;
         extensionButton.OnClick = onClick;
+        return AddButton(extensionButton);
+    }
+
+    public static ExtensionButton AddButton(ExtensionButton extensionButton)
+    {
         buttons.Add(extensionButton);
         return extensionButton;
     }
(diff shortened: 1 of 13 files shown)


monkeymanboy:
Right now there's not a lot of plugins so it's not a huge concern, but it would probably be worth also making this right side pop out panel scroll in the future.

Assets/__Scripts/MapEditor/UI/Extensions/ExtensionButtons.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Events;
+
+public static class ExtensionButtons
+{
+    private static List<ExtensionButton> buttons = new List<ExtensionButton>();
+
+    public static ExtensionButton AddButton(Sprite icon, string tooltip, UnityAction onClick)
```
Top-Cat, line 10: Why supply these properties to this function rather than an ExtensionButton itself?
Seems like it would be more future proof to use a builder-esque pattern where you construct the button and then supply it so old plugins don't break if you add a new property for instance.

Top-Cat:
Looks pretty cool, would make a lot of sense to put the error checker here rather than running its button around the top of the screen to avoid ui changes

Assets/__Scripts/MapEditor/UI/Extensions/ExtensionButtons.cs
monkeymanboy: Main idea was to keep it simple for the majority of use cases since those properties are the only 3 most plugins will touch. This way most of them can even just ignore the returned ExtensionButton since they'll never use it. Having another AddButton that just takes an ExtensionButton would be a good idea though.

Caeden117: PR otherwise looks good, I'll wait for this to be resolved first before merging.

Caeden117 (approved):
Alright, this looks good to merge now.

PR #271, opened 2021-05-07
Make aalto happy

Top-Cat:
Maybe LockedOntoNoteGrid should be re-enabled when exiting lightshow mode. But this restores the behaviour before it was broken by changing the default value.

Also allowed fade/flash to be placed in all lights lane while in prop mode

### DIFF
--- a/Assets/__Scripts/UI/LightshowController.cs
+++ b/Assets/__Scripts/UI/LightshowController.cs
@@ -7,19 +7,31 @@ public class LightshowController : MonoBehaviour, CMInput.ILightshowActions
     [SerializeField] private CameraController cameraController;
 
     private bool showObjects = true;
+    private bool previouslyLocked;
 
-    public void UpdateLightshow(bool enabled)
+    public void UpdateLightshow(bool enable)
     {
-        showObjects = enabled;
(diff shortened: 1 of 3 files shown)


Assets/__Scripts/UI/LightshowController.cs
```diff
@@ -18,10 +18,7 @@ public void OnToggleLightshowMode(InputAction.CallbackContext context)
     {
         if (context.performed)
         {
-            if (cameraController.transform.parent != null)
-            {
-                cameraController.OnAttachtoNoteGrid(context);
-            }
+            cameraController.LockedOntoNoteGrid = false;
```
Caeden117, line 21: Could we have it remember the previous state, and re-apply that when lightshow mode is off? It would be a QOL thing mainly for 360/90 mode

Caeden117 (changes requested):

Assets/__Scripts/UI/LightshowController.cs
Top-Cat: We can, just as per the original description this is how it was before.

Top-Cat:
Way more painful than expected

Caeden117 (approved):
pain is sacrifice to make aalto happy

PR #273, opened 2021-05-09
Added Colour Toggle Keybind

XAce1337manX:
Don't think anyone would mind moving the rotate in place keybinds to Alt + Q/E.

### DIFF
--- a/Assets/__Scripts/MapEditor/UI/KeybindUpdateUIController.cs
+++ b/Assets/__Scripts/MapEditor/UI/KeybindUpdateUIController.cs
@@ -16,6 +16,8 @@ public class KeybindUpdateUIController : MonoBehaviour, CMInput.IWorkflowsAction
     [SerializeField] private Toggle blueToggle;
     [SerializeField] private GameObject precisionRotationContainer;
 
+    private bool redSelected = true;
+
     void Awake()
     {
         UpdatePrecisionRotationGameObjectState();
@@ -44,6 +46,7 @@ public class KeybindUpdateUIController : MonoBehaviour, CMInput.IWorkflowsAction
     {
         if (!context.performed) return;
         blueToggle.onValueChanged.Invoke(true);
+        redSelected = false;
         placeMode.SetMode(PlacementModeController.PlacementMode.NOTE);
         lightMode.UpdateValue();
     }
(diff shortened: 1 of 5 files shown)


Assets/__Scripts/MapEditor/UI/KeybindUpdateUIController.cs
```diff
@@ -52,10 +55,26 @@ public void OnPlaceRedNoteorEvent(InputAction.CallbackContext context)
...
         redToggle.onValueChanged.Invoke(true);
+        redSelected = true;
         placeMode.SetMode(PlacementModeController.PlacementMode.NOTE);
         lightMode.UpdateValue();
     }
 
+    public void OnToggleNoteorEvent(InputAction.CallbackContext context)
+    {
+        if (!context.performed) return;
+        if (redSelected)
```
Top-Cat, line 66: This should probably be checking the current `isOn` value of `blueToggle` or `redToggle` rather than trying to duplicate that state in another variable.

Top-Cat: Or, having looked at it, we seem to not currently be setting the toggles themselves. So either we can do that in `ColorTypeController` or check `leftSelected.enabled` which is being set to change the visual style.

Caeden117:
> Don't think anyone would mind moving the rotate in place keybinds to Alt + Q/E.

Being one out of.... three 360 mappers in existence, i kinda do?

XAce1337manX:
So leave rotate in place keybinds as Q/E and have the toggle as Alt + Q? Or just have the toggle as a different keybind.

Top-Cat:
As they're just defaults you could make the toggle Alt+Q and people can swap them around if they want.
I don't know if there's any other keys that people would be happy with. Do we have anything bound to tab?

XAce1337manX:
Tab is bound to the right panel. (Chroma/Strobe generator/waveform)

Caeden117:
Yeah lets have the default be Alt-Q for now

Assets/__Scripts/MapEditor/UI/KeybindUpdateUIController.cs
XAce1337manX: Checked `leftSelected.enabled`.

PR #276, opened 2021-05-14
Song List UI

Top-Cat:
image

Incorporated most of the useful features that have been implemented previously/suggested

- More metadata shown for each song (similar display to in game) Had to try really hard to get this to not lock up for long periods, it can still hitch a little bit but only when scrolling a lot
- No more paging
- Search highlights the matching sections
- Implemented favouriting, stores a small hidden file in the song directory rather than saving to a separate config
- Different sort modes using Caeden's favourite collection type


Assets/__Scripts/BeatSaberSong.cs
```diff
@@ -220,6 +221,30 @@ public JSONNode ToJSONNode()
...
     }
 
+    public DateTime LastWriteTime;
+    private bool _isFavourite;
+    public bool IsFavourite
+    {
+        get => _isFavourite;
+        set
+        {
+            var path = Path.Combine(directory, ".favourite");
```
Caeden117, line 231: May I ask why we're using the filesystem for tracking favorites? If cross-editor compatibility is not an issue then we can also consider utilizing the `editors` object and store it there.

Caeden117, line 231: oi this is quite bri'ish of you, innit?

Top-Cat: Using the editors option was kinda second on the list.
The main reason I didn't do that is that currently we always save the diffs as well as the Info.dat and on bigger maps that's really slow. It would also mean that maps downloaded off beatsaver would randomly be favourited while the hidden file should not get packaged. And, as you pointed out, is is something that MMA2 could use too.

Assets/__Scripts/UI/SongSelectMenu/SongList.cs
```diff
@@ -1,77 +1,137 @@
...
+        var iterBeginTime = Time.realtimeSinceStartup;
         foreach (var dir in directories)
         {
-            BeatSaberSong song = BeatSaberSong.GetSongFromFolder(dir);
+            if (Time.realtimeSinceStartup - iterBeginTime > 0.02f)
+            {
+                UpdateSongList();
+                yield return null;
+                iterBeginTime = Time.realtimeSinceStartup;
+            }
```
Caeden117, line 132: ooc, what's the purpose of this bit?

Assets/__Scripts/BeatSaberSong.cs
Top-Cat: Time to rename to `.fav` or `.star` to avoid this debate 👀

Caeden117: brb gonna get me some tea n' crumpets

~~i honestly dont care, just like poking fun~~

Assets/__Scripts/UI/SongSelectMenu/SongList.cs
Top-Cat: To prevent stalling the main thread, I have >400 maps in custom levels and that takes more than 1 frame to load.
But if we yield every loop then it takes >400 frames to load, this tries to find a middle ground.

I guess the alternative is to put it all on another thread and then poll every frame until it completes... 🤔

Caeden117: nothing needs to be done, i was just wondering the purpose behind it

Assets/__Scripts/UI/SongSelectMenu/SongListItem.cs
```diff
@@ -1,27 +1,341 @@
...
+    private static JSONObject _songCoreCache;
+
+    private bool _ignoreToggle = false;
+
+    private void Start()
+    {
+        rightPanel.SetActive(false);
+        _bg = GetComponent<Image>();
+        // I have sinned
+        _songList = FindObjectOfType<SongList>();
```
Caeden117, line 48: excuse me what the fuck is this

image

SkylerWallace:
unknown-8
My contribution.  Took a lot of your ideas and made the appearance more consistent in terms of coloring, padding, and spacing.  Can provide SVG's for icons.

Top-Cat:
I'll quickly try to list out what changes are relevant which is hard as your suggestion is just a screenshot.

- Swapped tabs / new map + refresh buttons
Don't really see that this is better or worse really, perhaps it's to allow room for the folder button (will get to that) but there's not even space in there for a third tab.
- Folder button
Seen a couple suggestions for this but I don't see why it should be part of this PR
- Font in the list view
As above I think that changing the main font is outside of the scope of this PR. In your image the folder name text is especially small. I had to fiddle a lot with TMP to get the text in this PR to be readable at all. At least without increasing the canvas resolutions (maybe we should) I don't know how to improve the clarity and wouldn't be comfortable making it that small
- Lower line heights
This just seems unnecessary. I guess the implication is you thought that the view above was confusing to see what name + artist was attributed to which song and I don't really see the problem and I think you've taken it too far if anything.
- Shortened folder names
I can see why you'd do this as if the parent folder is too long all the lines end up being the same. I do think the full path is going to reduce confusion for new users who I see regularly saying that files are in the right place when they aren't. I'd think that scrolling text or something similar would do a better job at solving that problem. Again, something for the future not this PR.
- Time/BPM icons
I can't understand from the image how you intend to handle songs longer than 10 minutes. Bit of a nitpick I guess but the current design was modeled after the in game list in order to try and make it feel more familiar.
image

If there's anything else you changed that you'd like me to consider please feel free to explain what it is and why it's good.
Thanks :)

SkylerWallace:
> **Swapped tabs / new map + refresh buttons**
Don't really see that this is better or worse really, perhaps it's to allow room for the folder button (will get to that) but there's not even space in there for a third tab.

I initially made this change because it made more sense to me, but coincidentally it also vertically aligns the Refresh icon with the Sort icon so it creates a more consistent spacing of icons.  I'd suggest implementing a horizontal scroll in the Tab container when adding more than 2 tabs...alternatively what if the Song Select box was just larger and centered?  The right side of the Song Select menu is already pretty barren and currently only really serves the purpose of making space for a logo.

> **Folder button**
Seen a couple suggestions for this but I don't see why it should be part of this PR

Would be really nice to be able to add more folder locations, especially for us "power mappers".  Maybe can add this functionality in the future.

> **Font in the list view**
As above I think that changing the main font is outside of the scope of this PR. In your image the folder name text is especially small. I had to fiddle a lot with TMP to get the text in this PR to be readable at all. At least without increasing the canvas resolutions (maybe we should) I don't know how to improve the clarity and wouldn't be comfortable making it that small

I don't have whatever font CM is currently using, I just used Calibri for now.  I'd still suggest changing the system font anyways; MMB's original UI PR's did that and there wasn't really any issue with it.  Not sure what exactly the canvas resolution is, but if it's limiting what types of fonts can be used that is probably something that should be addressed as long as changing it doesn't cause too many adverse effects.

> **Lower line heights**
This just seems unnecessary. I guess the implication is you thought that the view above was confusing to see what name + artist was attributed to which song and I don't really see the problem and I think you've taken it too far if anything.

The line heights aren't really lowered as much as they're just vertically centered to the album art now.  The original design you posted makes it much more difficult to distinguish between individual songs; there's consistent spacing on every line, even when between individual songs.  As for how much spacing there should be, that can be tweaked until everybody is happy, but I still think it should be addressed as currently there's not really any padding between songs in the list view.

> **Shortened folder names**
I can see why you'd do this as if the parent folder is too long all the lines end up being the same. I do think the full path is going to reduce confusion for new users who I see regularly saying that files are in the right place when they aren't. I'd think that scrolling text or something similar would do a better job at solving that problem. Again, something for the future not this PR.

I did this just because I didn't want to put some arbitrary file path in the mockup.  Would definitely agree that a tickertape effect would be the best way to solve the issue you brought up.  I'd suggest truncating the file path and when the user hovers over a map, the tickertape effect occurs to display the entire file path.

> **Time/BPM icons**
I can't understand from the image how you intend to handle songs longer than 10 minutes. Bit of a nitpick I guess but the current design was modeled after the in game list in order to try and make it feel more familiar.

I can definitely respect that decision.  I do slightly prefer the timer icon for consistency but it's by no means objectively better to do it that way.  To solve songs longer than 10 minutes, the cleanest solution would likely be just to swap the icon and text, then right-align the text.

Caeden117:
Alright so I've been locally using this for a little bit, here's my thoughts on it:

On the topic of Skyler's earlier mockup:
- I do like Skyler's idea of swapping location of the Custom/WIP Levels tabs and the New Song/Refresh buttons, especially if we plan on introducing more song folders in a future PR
- I do think the padding and spacing there is a bit better, primarily with the song list items
- The "New Map" button is more pronounced in his mock-up, which might help people who are just starting out but I'm not quite sure

I did encounter a few bugs with this PR that I've fixed locally. Namely, the modifications made to `SongInfoEditUI.cs` did not take newly created maps into account. Locally, I've added a quick null check before setting the song duration, as well as moving that whole block of code to execute after saving the song (Ensuring that the directory is created). I'm not sure if this was a Unity moment or not, but the references to the Custom/WIP Tab images in the `SongList` component appeared to be missing. Either way, I re-assigned the images.

My only other nitpick currently is that there aren't any tool-tips on the Filter By button, even if the icons are pretty self explanatory i feel like some localized tooltips would be nice

PR #289, opened 2021-06-02
Fast Raycasting

Caeden117:
Figured I should start making my own branches and pull requests instead of pushing everything to `dev`, for the purposes of code review. This PR will probably be merged after the existing ones.

This PR introduces a new system to handle raycasts that completely circumvents Unity's physics system.

To keep things simple, the only use that Unity Physics saw in ChroMapper was to handle 3D raycasting from the mouse into the scene. This includes stuff like placing objects, alt-clicking and dragging objects, and selecting objects. On larger maps, Unity Physics added about 8ms of CPU overhead _every frame_, and that is just to resync the objects to the Physics engine. For context, in order to achieve 60 FPS, the entire frame needs to be rendered in less than 16ms.

To solve this, I've written my own custom Raycasting system from scratch. This Raycasting system does not require any resyncing, and uses a fast Möller–Trumbore ray-triangle intersection algorithm to handle collisions. In layman's terms, instead of 8ms being consumed every frame by Unity Physics, this custom Raycasting system consumes about 1-2ms *at most*.

In *theory*, this PR should be nothing but free FPS gains for everyone. However, in testing there are a few slight edge cases that need ironed out, so consider this PR a WIP.

To keep things fast, all objects use extremely low-poly meshes to perform collision detection on. In most cases, it will be the standard Unity cube. Currently, the only exception to this are bombs, which use its own mesh for collision. With ~200 triangles in the Bomb mesh, I'm not *super* worried unless bomb mapping becomes extremely prevalent.

**BEFORE** (Notice the constant orange consuming about 8ms of frametime)

**AFTER**


Caeden117:
So DaNike ~~and I~~ came up with a method for raycasting that is incredibly optimized.
image
*(For context, that is on Bury A Friend)*

TL;DR: We render an additional pass each frame, which renders collision meshes into a separate render texture. On request, we look up the pixel under our mouse, and immediately and accurately get the object.

This does add a little bit of GPU overhead each frame, which is a trade off I am willing to take. Currently, the lack of GPU Instancing across ChroMapper does leave some room for improvement on the GPU side, and also legitimately interferes with the new raycast approach a little bit. As such, consider this PR as blocked until the [`gpu-instancing`](https://github.com/Caeden117/ChroMapper/tree/gpu-instancing) branch is complete and merged.

Caeden117:
Closed as we're using a modified version of the older (not GPU raycasted) approach, in commit 8001a4cc9207852746215dcd3d2d03664f2d6380

PR #294, opened 2021-06-30
Bug fixes

Top-Cat:
- Don't remove _active if it's false, contrary to the docs this actually does something
See: https://discord.com/channels/702230714585710602/702232127755780197/856579355899199530
- Fix boost event materials not updating [CM-70]
- Avoid error when writing duration cache if it doesn't exist yet
- Change lightid filtering in gradient generator, add tests
See: https://discord.com/channels/702230714585710602/702232127755780197/857335694723907585
- Cancel previous song loads when triggering a new one
See: https://discord.com/channels/702230714585710602/702232127755780197/857429175191470141


Caeden117 (approved):
k

PR #295, opened 2021-07-01
Add functionality for exporting to Quest

Fernthedev:
This adds a button for exporting to Quest devices. There are some caveats I would like to address later on (with opinions on how to address them)
- This requires adb.exe to be in the system path. I would like to be able to add functionality to download adb and extract it automatically rather than bundling it due to licensing.
- This does not check if Quest device is connected or not
- This does not validate if the files copied successfully
- This requires the Quest device to have RuntimeSongLoader which is currently not in public release state. (though it is stable and public). Once BMBF releases, this will be a coremod and a non-issue.


Assets/Locales/SongEditMenu_es.asset
```diff
@@ -368,5 +368,13 @@ MonoBehaviour:
       it instead?
     m_Metadata:
       m_Items: []
+  - m_Id: 78
+    m_Localized: "Archivos exportados con \xE9xito a Quest"
```
Top-Cat, line 372: As helpful as this is the way we have crowdin set up this won't get uploaded and will have to be entered there manually anyway

Assets/__Scripts/Singletons/Adb.cs
```diff
@@ -0,0 +1,165 @@
...
+                return Path.GetFullPath(fileName);
+
+            var values = Environment.GetEnvironmentVariable("PATH");
+
+            return values?.Split(Path.PathSeparator).Select(path => Path.Combine(path, fileName)).FirstOrDefault(File.Exists);
+        }
+
+        public static void Initialize(string adbPath = "adb.exe")
+        {
+            Assert.IsTrue(ExistsOnPath(adbPath),
```
Top-Cat, line 56: Not keen on this being uncaught. Why are we not showing a dialog?

Assets/__Scripts/Singletons/Adb.cs
```diff
@@ -0,0 +1,165 @@
...
+        {
+            ValidateADB();
+
+            string makeParentsFlag = makeParents ? "-p" : "";
+
+            _process.StartInfo.Arguments = $"shell mkdir {devicePath} {makeParentsFlag} -m {permission}";
+
+
+
+            return await Task.Run(() =>
```
Top-Cat, line 107: Can't help but notice this block is in triplicate

Assets/__Scripts/UI/SongInfoEditUI.cs
```diff
@@ -369,62 +371,117 @@ private void HandleDeleteMap(int result)
...
+
+    // TODO: Move this to a proper location or make configurable with a default
+    private const string QUEST_CUSTOM_SONGS_LOCATION = "sdcard/ModData/com.beatgames.beatsaber/Mods/SongLoader/CustomLevels";
+
+    /// <summary>
+    /// Exports the files to the Quest using adb
+    /// </summary>
+    public async void ExportToQuest()
+    {
+        var songExportPath = Path.Combine(QUEST_CUSTOM_SONGS_LOCATION, Song.cleanSongName).Replace("\\", @"/");
```
Top-Cat, line 435: Not sure why this needs to be a string literal

Top-Cat:
As you mentioned in the description it would be ideal if we checked a quest was connected and only showed the option if it's detected otherwise I just have a button that throws errors. Good start though :)

Fernthedev:
> As you mentioned in the description it would be ideal if we checked a quest was connected and only showed the option if it's detected otherwise I just have a button that throws errors. Good start though :)

That was definitely my intention, though I've no idea how you'd think it should be done. Let me know if you have any suggestions for hiding/showing the button when a Quest is connected.

As for the assert, that was just for me to test, I'll have to add a dialogue for it somehow.

Top-Cat:
> That was definitely my intention, though I've no idea how you'd think it should be done. Let me know if you have any suggestions for hiding/showing the button when a Quest is connected.

I don't own a quest so I don't know that I'm the best person to suggest answers but at a minimum check that `adb devices` returns something? That doesn't guarantee the device isn't a phone, you could go further and run other adb commands.

I guess it's possible you have connected it but haven't enabled debugging? We may just need to add some documentation around that rather than trying to go above and beyond to detect a quest that won't play ball.

Fernthedev:
I was actually referring to the UI code of it. I already have a plan for checking if a Quest is connected, though not 100% sure if it will work. Maybe I'll look into SideQuest's source

Top-Cat:
> I was actually referring to the UI code of it. I already have a plan for checking if a Quest is connected, though not 100% sure if it will work. Maybe I'll look into SideQuest's source

Well, either hide or disable the button until you've detected a quest. The horizontal layout should reflow if you hide it and I'd say as long as you're not re-hiding the button every time that screen loads it's probably better to just hide it completely for normies. (You could recheck every time but keep the state from the previous check until you're done)

Fernthedev:
For now, the button will show/hide when needed. However, the assert is still there but the exception is handled. (depends on your definition of "handled")

There is now the problem that Quest functionality is not obviously documented and no "one click setup" for ADB. I'd be keen on opinions on how to handle these two problems.

Top-Cat:
LogError will still cause the console to appear which is just as annoying as a dialog box. So if you're not comfortable with a dialogue box you shouldn't be comfortable with LogError

We can add documentation later (or you can make an account and write some yourself) on the CM wiki https://chromapper.atlassian.net/wiki

It seems that normally adb install is left to the user. Is it not included in the quest software?
Of course that doesn't guarantee the exe is on the PATH. However there are a few libraries for talking directly to the adb server rather than starting and parsing the cli.

Fernthedev:
> It seems that normally adb install is left to the user. Is it not included in the quest software?
It is not, SideQuest (I assume anyways) downloads it's own and uses it. ADB is not specific for Quest either so a user can technically already have it installed if they use something along the lines of Android Studio for example.

> Of course that doesn't guarantee the exe is on the PATH. However there are a few libraries for talking directly to the adb server rather than starting and parsing the cli.

I used my own because I didn't want to add extra dependencies and because those other libraries undoubtedly also require ADB installed anyways (unless somehow they don't)

Top-Cat:
> I used my own because I didn't want to add extra dependencies and because those other libraries undoubtedly also require ADB installed anyways (unless somehow they don't)

They require an adb server to be running, they do not require the adb binary. That's the point I'm trying to make.
I'm dubious that sidequest is doing anything different to that.

Fernthedev:
It seems you are correct, I misunderstood how ADB works. SideQuest has a NodeJS implementation, however there is a .NET implementation of the `adb` protocol here: https://github.com/quamotion/madb
As noted in the .NET implementation, it still does require [ADB to be installed in the first place](https://github.com/quamotion/madb#starting-the-adb-server)

For reference, SideQuest seems to use [this](https://github.com/DeviceFarmer/adbkit) however this ALSO requires `adb.exe` to be installed since it is not a server implementation. Feel free to suggest one though if you do find it.

Top-Cat:
I think that .NET library was the one I found with a quick google too, if it works then I don't see why not use it.

Fernthedev:
So I know I've been rather cyclic about this but the parsing does work, though if you insist on it (because it's probably more robust, trustworthy or performant) then I'll switch to it. 

Just to reinforce my understanding of SideQuest, it does seem that it downloads ADB. References:
- [Finding adb](https://github.com/SideQuestVR/SideQuest/blob/d278c3bf7d4cea7f452342fd9bf4bd3a008f181c/desktop-app/src/app/adb-client.service.ts#L60)
- [Download here](https://github.com/SideQuestVR/SideQuest/blob/22845fbca322e904420ecf9942686488ff86fbc0/electron/platform-tools.ts#L6-L19)

Top-Cat:
Interesting

I'd say if we're downloading adb then stick with this approach as we should be able to guarantee it will work as we downloaded the exe and put it there.
If we let the user install adb in any way use a library that doesn't care where the binary is.

Fernthedev:
ADB download dialogue has been implemented, though I suggest we get both Mac and Linux users to test it just to be 100% sure. The dialogue would preferably at first time or less intrusive to avoid PC only users complaining about it. Other than that, I have a feeling this will soon be ready to get properly reviewed.

Assets/__Scripts/Singletons/Adb.cs
```diff
@@ -0,0 +1,293 @@
...
+                case RuntimePlatform.OSXPlayer:
+                    return PLATFORM_TOOLS_DOWNLOAD_GENERIC + "darwin.zip";
+                case RuntimePlatform.WindowsPlayer:
+                case RuntimePlatform.WindowsEditor:
+                    return PLATFORM_TOOLS_DOWNLOAD_GENERIC + "windows.zip";
+                case RuntimePlatform.LinuxPlayer:
+                case RuntimePlatform.LinuxEditor:
+                    return PLATFORM_TOOLS_DOWNLOAD_GENERIC + "linux.zip";
+                default:
+                    throw new InvalidOperationException("How could this even happen?");
```
Top-Cat, line 72: Yeah... who would try and run this on android.... oh wait
Android

On a more serious note it may be better to use compiler macros so that this doesn't have to be checked at runtime.

Fernthedev: That would be lovely, how would I go about doing that with Unity? Also I was funnily enough going to add Android support just for the joke (especially since Quest is Android, mapping in VR ???)

Top-Cat: https://github.com/Caeden117/ChroMapper/blob/41a64f50212de47b252a7d33881cfab8f78aea32/Assets/StandaloneFileBrowser/StandaloneFileBrowser.cs#L18

Is probably the most complete example although we have a couple other checks just for windows like:

https://github.com/Caeden117/ChroMapper/blob/41a64f50212de47b252a7d33881cfab8f78aea32/Assets/__Scripts/UI/Contributors/ContributorListItem.cs#L90

Assets/__Scripts/UI/SongInfoEditUI.cs
```diff
@@ -124,12 +131,151 @@ public static int GetDirectionalEnvironmentIDFromString(string platforms)
...
+        questExportButton.SetActive(false);
+
+        try
+        {
+            Adb.Initialize();
+            StartCoroutine(CheckIfQuestIsConnected());
+        }
+        catch (AssertionException)
+        {
+            Debug.LogError("No ADB found, asking to download");
```
Top-Cat, line 143: REEEEEEEEEEEEE

Fernthedev:
So this PR is as I see ready to prepare for merging. A few things I'd like to mention
- UI code is horrible
- For testing, make sure you don't have `adb.exe` in your path and you can delete `quest-utils` in your Appdata folder (I forget where) or in your project directory if running through Unity
- Test on other OS' such as Mac and Linux. Windows has been tested.

ChroMapper.sln
```diff
@@ -1,59 +1,45 @@
 ﻿
-Microsoft Visual Studio Solution File, Format Version 12.00
```
Top-Cat, line 2: Why?

Assets/__Scripts/UI/FirstBootMenu.cs
```diff
@@ -74,8 +78,22 @@ public class FirstBootMenu : MonoBehaviour {
         }
 
         directoryCanvas.SetActive(true);
+
+        if (!Adb.IsAdbInstalled(out _))
+        {
+            PersistentUI.Instance.ShowDialogBox("Options", "quest.adb_not_found", result =>
```
Top-Cat, line 84: Personally would just show a button for this

Assets/__Scripts/TaskExtensions.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace __Scripts
```
Top-Cat, line 6: None of the files (at least in __Scripts) should have namespace afaik. Not even sure this is the right namespace.

Assets/__Scripts/UI/SongInfoEditUI.cs
```diff
@@ -124,12 +132,106 @@ public static int GetDirectionalEnvironmentIDFromString(string platforms)
...
     }
 
+    void OnDestroy()
+    {
+        if (questCandidates.Count <= 0) return;
+        Debug.Log("No longer checking for quest");
+        Adb.Dispose().ConfigureAwait(false);
+    }
+
+
```
Top-Cat, line 168: You have entered the void.

Assets/__Scripts/UI/SongInfoEditUI.cs
Top-Cat: *cough*

JD-The-65th:
This PR will be here til the end of time lol

Assets/__Scripts/Singletons/AdbUI.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            void UpdateLabel(bool extracting)
+            {
+                progressBarComponent.WithCustomLabelFormatter(f =>
+                {
+                    return LocalizationSettings.StringDatabase.GetLocalizedString("Options",
+                        extracting ? "quest.extracting_download" : "quest.downloading_progress", new object[] { f * 100 });
+                });
+            }
```
Caeden117, line 53: Is there any real purpose to these inlined methods? It seems like a code smell to me, especially when `Fail` is only effectively called once, and `UpdateLabel` can easily be abstracted away altogether.

Assets/__Scripts/UI/Options/AdbHandler.cs
```diff
@@ -0,0 +1,45 @@
+﻿using QuestDumper;
+using UnityEngine;
+
+namespace __Scripts.UI.Options
```
Caeden117, line 4: Classes in `__Scripts` don't have a namespace (mostly a Unity moment). This should be removed for consistencies' sake.

Assets/__Scripts/UI/FirstBootMenu.cs
```diff
@@ -65,6 +66,9 @@ private void Start()
         directoryCanvas.SetActive(true);
     }
 
+    public void InstallAdb() => StartCoroutine(AdbUI.DoDownload());
```
Caeden117, line 69: If the dialog box inside `AdbUI.DoDownload()` is unclosable by design, this should _really_ be wrapped in another "Are you sure" box, or offer the ability to cancel the ADB download.

I'm not going to have people click this button (either unintentionally, out of curiosity, etc.) and have no effective way to prevent or cancel the download without closing ChroMapper entirely. It would also be smart to expose UI for uninstalling ADB. Not all ChroMapper users are Quest users.

Assets/__Scripts/UI/Options/AdbHandler.cs
```diff
@@ -0,0 +1,45 @@
...
+        public void ToggleADB()
+        {
+            if (!Adb.IsAdbInstalled(out _))
+            {
+                StartCoroutine(AdbUI.DoDownload());
+            }
+            else
+            {
+                _betterToggle.IsOn = true;
+            }
```
Caeden117, line 42: See previous comment on unclosable dialog box

Caeden117 (changes requested):
Hi, sexy jaroslav beck here, today i'm going to crush your dreams

Assets/__Scripts/UI/FirstBootMenu.cs
Fernthedev: Realistically they only lost a few MB of storage (30MB on Windows) and it runs ADB server in the background when editing song metadata. Wouldn't say PC users would even notice or care unless you tell them. I was planning on adding uninstall anyways somehow, though I'm probably too lazy to add a cancel method or "are you sure " dialog since that means more spaghetti for me. The download is also about 10MB which for majority of users is too small to even cancel out of.

Assets/__Scripts/UI/Options/AdbHandler.cs
```diff
@@ -0,0 +1,51 @@
...
+    private void Start()
+    {
+        _betterToggle = GetComponent<BetterToggle>();
+        // Set toggle
+
+        
+        // Update UI ugh
+        toggledBySelf = true;
+        _betterToggle.SetUiOn(Adb.IsAdbInstalled(out _));
+        toggledBySelf = false;
```
Caeden117, line 23: If you want to clean this up, I'd throw out the `BetterToggle` entirely and use a regular `Toggle` (can use the CMUI prefab as a base), then you can remove this whole `toggledBySelf` bullshit and instead use `Toggle.SetValueWithoutNotify()`. Alternatively, you can just replicate that method in `BetterToggle`.

Caeden117 (approved):

PR #296, opened 2021-07-04
Be lazier loading diffs

Top-Cat:
I realised that I was constructing a Lazy and then immediately calling a method that tries to access its value. The idea was that even though we need to load the diff file to get the new env enhancements that we wouldn't load diffs unless we needed to display this data.

Changed it so that instead a null value is used internally to denote no change has been made. The map will be loaded and value populated when the diff is selected (this could further be optimised so that it only loads when showing the enhancements tab).

Also changed the "Open Editor" button so that it pulls from the lazy value.

Combined this means that for a full spread instead of loading diff files 6 (every diff on screen + one of them twice) times before you get to the editor we now load 1 time (or as many as you click on).


Caeden117 (approved):
edit screen loading times back to nyooom

PR #301, opened 2021-07-14
Raycasting fixes

Top-Cat:
I was having some problems with selecting notes and events.

There are three modes depending on camera direction sadly the forward and backwards modes don't consider high FOV values so objects at the edge of the window that are "behind" the camera can't be selected.

The alternating mode works reasonably well in the middle of a map but at the start it would stop processing as soon as it considered a chunk before 0 so if the camera is over chunk 0 it checks 0, 1, -1, 2 -> stop. Because alternating has reached -1 objects in 3 or higher are inaccessible.

### In this PR
Just removed the forward/backward modes as high fov sadly makes them useless. We always alternate.

Code for getting the next group id simplified (I can't even read the old code)

We always do enough loops to reach the end in the furthest direction and skip iterations that are outside the bounds (possibly could just remove this check as it should fall through the trygetvalue afterward). lowestKey and highestKey are now accurate for loaded containers as when deregistering unused keys are removed.

### DIFF
--- a/Assets/__Scripts/Map/Obstacles/BeatmapObstacleContainer.cs
+++ b/Assets/__Scripts/Map/Obstacles/BeatmapObstacleContainer.cs
@@ -111,29 +111,6 @@ public class BeatmapObstacleContainer : BeatmapObjectContainer
             transform.RotateAround(rectWorldPos, transform.forward, localRotation.z);
         }
 
-        UpdateCollisionGroupsWithDuration(duration);
-    }
-
-    private void UpdateCollisionGroupsWithDuration(float duration)
-    {
-        var chunkStart = ChunkID;
-
-        var chunkEnd = (int)((obstacleData._time + duration) / Intersections.ChunkSize);
-
-        if (chunkStart > chunkEnd)
-        {
-            // me and the boys casually flipping shit
-            (chunkStart, chunkEnd) = (chunkEnd, chunkStart);
-        }
-
-        var range = Enumerable.Range(chunkStart, Mathf.Max(chunkEnd - chunkStart, 1));
-
-        foreach (var collider in colliders)
-        {
-            var unregistered = Intersections.UnregisterColliderFromGroups(collider);
-            collider.CollisionGroups.Clear();
-            collider.CollisionGroups.AddRange(range);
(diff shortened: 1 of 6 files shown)


Top-Cat:
I've now noticed this still isn't perfect because with the camera placed above beat 1 objects placed in beat 1 will always be selected/deselected first (the wall behind).

reeeee

Top-Cat:
I've had to remove a few optimisations that were causing issues in corner cases but I've also added an optimisation that means only one raycast is done per frame for selection raycasts.

Specifically:
- We can't stop checking mesh tris after finding one that hits as we use the intersection to find the distance and if you start with the furthest tri sometimes and nearest other times then overlapping meshes can be incorrectly sorted.
- We can't stop checking chunks after finding a hit in the current one (see image in previous comment)

The slowest I've managed to time the process at is ~20ms, but only on a map that is basically makes CM crawl anyway. Normal maps are <5ms

Assets/__Scripts/Map/Obstacles/BeatmapObstacleContainer.cs
```diff
@@ -111,29 +111,6 @@ public override void UpdateGridPosition()
...
-
-        var range = Enumerable.Range(chunkStart, Mathf.Max(chunkEnd - chunkStart, 1));
-
-        foreach (var collider in colliders)
-        {
-            var unregistered = Intersections.UnregisterColliderFromGroups(collider);
-            collider.CollisionGroups.Clear();
-            collider.CollisionGroups.AddRange(range);
-            if (unregistered) Intersections.RegisterColliderToGroups(collider);
-        }
```
Caeden117, line 137: Wouldn't this removal affect selecting objects that are behind long obstacles? What about negative duration walls?

Top-Cat: Because of the change where we don't stop after checking a single chunk but check all the loaded chunks anything on screen should get checked so adding the obstacle to multiple chunks isn't necessary.

Top-Cat: And those cases were broken before this change which is exactly why I've changed it. This system didn't work.

PR #305, opened 2021-07-22
RTFM

Top-Cat:
Reading the manual is often useful


Caeden117 (approved):
Yeah jeez top cat, RTFM

~~Just going to casually ignore the fact that I would've made the same exact mistake~~

PR #308, opened 2021-07-23
Make options searchable

Top-Cat:
All the options have a custom set of keywords which are mostly just their text but I've added a few extra to make some options easier to find. Users can eventually feedback things they expected to work that don't.

search1
search2


Assets/__Scripts/UI/Options/Plugins/PluginInfoContainer.cs
```diff
@@ -6,9 +6,14 @@ public class PluginInfoContainer : MonoBehaviour
     [SerializeField] private TextMeshProUGUI nameText;
     [SerializeField] private TextMeshProUGUI versionText;
 
+    public SearchableSection SearchableSection;
+    [SerializeField] private SearchableOption searchableOption;
```
Caeden117, line 10: Is the public `SearchableSection` necessary here?

Caeden117 (approved):
code LGTM, will locally test before merging

PR #309, opened 2021-07-23
add keybind to toggle-mute hitsounds

GalaxyMaster2:
Adds a keybind to mute/unmute hitsounds. Current (temporary) default binding is Alt+F1, but I'm open to suggestions.

Toggling hitsounds during playback will have a small delay due to hitsounds being scheduled ahead of time. This could potentially be solved by canceling and rescheduling all hitsounds, or by attempting to change their volume after scheduling, but the issue may be considered minor enough to not worry about.

### DIFF
--- a/Assets/__Scripts/MapEditor/Hit Sounds/HitSoundVolumeController.cs
+++ b/Assets/__Scripts/MapEditor/Hit Sounds/HitSoundVolumeController.cs
@@ -1,0 +1,26 @@
+﻿using UnityEngine;
+using UnityEngine.InputSystem;
+
+namespace Assets.__Scripts.MapEditor.Hit_Sounds
+{
+    class HitSoundVolumeController : MonoBehaviour, CMInput.IAudioActions
+    {
+        [SerializeField] private float lastVolume = 0f;
+
+        public void OnToggleHitsoundMute(InputAction.CallbackContext context)
+        {
+            if (!context.performed) return;
+
+            float currentVolume = Settings.Instance.NoteHitVolume;
+            if (currentVolume == 0f)
+            {
+                Settings.Instance.NoteHitVolume = lastVolume;
+            }
+            else
+            {
+                lastVolume = currentVolume;
+                Settings.Instance.NoteHitVolume = 0f;
+            }
+        }
+    }
+}
--- a/Assets/__Scripts/MapEditor/Hit Sounds/HitSoundVolumeController.cs
+++ b/Assets/__Scripts/MapEditor/Hit Sounds/HitSoundVolumeController.cs
@@ -20,4 +20,13 @@ public class HitSoundVolumeController : MonoBehaviour, CMInput.IAudioActions
             Settings.Instance.NoteHitVolume = 0f;
         }
     }
+
+    private void OnDestroy()
+    {
+        float currentVolume = Settings.Instance.NoteHitVolume;
+        if (currentVolume == 0f)
+        {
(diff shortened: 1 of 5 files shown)


Assets/__Scripts/MapEditor/Hit Sounds/HitSoundVolumeController.cs
```diff
@@ -0,0 +1,26 @@
+﻿using UnityEngine;
+using UnityEngine.InputSystem;
+
+namespace Assets.__Scripts.MapEditor.Hit_Sounds
```
Caeden117, line 4: Looks like this was added in the IDE, not through Unity

To keep consistency with the rest of the project, would you mind removing the namespace? That's about the only I'd like changed with this PR

Top-Cat:
So if you use this to mute hitsounds and quit the editor we're going to save the volume as 0 and when you relaunch you'll have to go into settings to put your volume back. This sounds like something that's going to get annoying.

Easiest solution might be to have this hook OnDestroy and put the volume back assuming that gets called before the save routine. Otherwise you'd have to separate this out from the existing setting.

Assets/__Scripts/MapEditor/Hit Sounds/HitSoundVolumeController.cs
GalaxyMaster2: Indeed it was, will do

GalaxyMaster2:
> So if you use this to mute hitsounds and quit the editor we're going to save the volume as 0 and when you relaunch you'll have to go into settings to put your volume back. This sounds like something that's going to get annoying.
> 
> Easiest solution might be to have this hook OnDestroy and put the volume back assuming that gets called before the save routine. Otherwise you'd have to separate this out from the existing setting.

Hooking OnDestroy works but still has some downsides. If I always apply the saved volume you could have your volume overwritten unexpectedly in the following scenario:
- Volume is at x value
- Mute keybind is pressed
- Volume is manually changed to y value
- User exits map
- Volume is set to x value unexpectedly

If I only apply the saved volume if the current volume is 0, you could still have unexpected behavior:
- Volume is at x value
- Mute keybind is pressed
- Volume is manually changed to a value >0
- Volume is later manually changed to 0
- User exits map
- Volume is set to x value unexpectedly

I think that last scenario is pretty unlikely to occur and I don't think it would be much of a problem.

Top-Cat:
The only solution that solves all of these is to have a separate bool that tracks this mute option and modify everywhere that currently uses the hit sound volume to be 0 when this is set.

It's not a huge deal either way as if you don't use the bind none of these senarios can happen but I'm sure someone will complain eventually.

GalaxyMaster2:
The nicest solution would probably be to add mute settings for each of the volumes, add mute buttons to each of the sliders in the options menu, and add additional mute keybinds for the other volume types. However that seems beyond my current abilities in Unity.

Adding a persistent setting not exposed in ui seems like a bad idea and could lead to major confusion. Resetting it to unmuted when exiting the editor would help, but could still lead to problems.

GalaxyMaster2:
I'm pretty sure this should solve all edge cases, and seems like the best solution next to adding proper mute settings.

Caeden117 (approved):
code LGTM, going to locally test before merging

unfortunately still have no good keybind ideas, unless you want to move/remove existing ones (`V` and finding a new place for the note cut "visualizers" could work)

PR #312, opened 2021-07-26
BPM change improvements

GalaxyMaster2:
Makes the cursor follow the grid on BPM change tweak (Alt+scroll), makes BPM change placement default to the previous BPM, and improves scrolling behavior around BPM changes.

About moving cursor on BPM change tweak:
- Only applies on BPM tweak
- Does not affect undo/redo
- Does not affect other means of editing BPM changes

The reasoning for this is as follows:
- The main use is in mapping BPM slides, where you are likely to use BPM tweak
- Always moving cursor when a BPM change is added or modified would be inconsistent with the behavior of map objects (notes, lights, etc.) which remain stationary
- Moving the cursor on undo/redo would require the above, unless a new action type was created specifically for BPM tweak
- Desired behavior of undo/redo becomes less clear when the cursor is moved before undo/redo


Caeden117 (approved):
LGTM

PR #313, opened 2021-07-26
remember song list sorting type

GalaxyMaster2:
Does what it says. Adds a new (hidden) setting that stores the last used sorting type.


Caeden117 (approved):
LGTM

PR #314, opened 2021-07-27
Fix rocket right lasers rotating the wrong way and fix panel rings not working in "_nameFilter" 

SuperSmay:
Changed the multiplier value on the rotating laser objects in Unity from 20 to -20 on the right rocket lasers so that they rotate in the correct direction, matching in game.
And added "Panels" to the name filter detection for the small rings


SuperSmay:
I hope I committed all the needed files, Unity created a bunch of extra crap and I might have missed something. If I did I can either resubmit or someone else can just go make the changes I did in about 3 minutes, or find a better way to do the same things

PR #316, opened 2021-07-28
Update Rocket Env

SuperSmay:
Changes to Rocket Environment: 
- Update laser angle to match in game angle
- Add the two missed center lights
- Add the missing construction below the environment
- Fix the right lasers rotating in the wrong direction


et --hard HEAD
```diff
@@ -0,0 +1,255 @@
...
+Author: Thomas Cheyney <top_cat@thomasc.co.uk>
+Date:   Mon Jul 12 10:48:39 2021 +0100
+
+    Improve options menu resilience
+
+[33mcommit d561e80ee0f72ef1a53139106d64826367060cff[m
+Author: Thomas Cheyney <top_cat@thomasc.co.uk>
+Date:   Mon Jul 12 10:42:23 2021 +0100
+
+    Fix selected tab when returning to song li
```
Caeden117, line 255: what is this file? i dont think its necessary here.

Top-Cat: Someone accidentally `git log > et --hard HEAD`

SuperSmay: Whoops. Not sure how I managed to accidentally do that but its been deleted

PR #317, opened 2021-07-29
Add transparent track backgrounds

SuperSmay:
Adds a slider to make track backgrounds transparent/invisible
This is done by essentially leaving the grids in "high contrast mode" by disabling the other method of high contrast mode and leaving the base alpha of the grid layer at 0. The base layer of the tracks is then set to either the normal or high contrast color, and if the transparency slider is set above 0%, the shader is replaced with a transparent unlit color shader from [here](https://gist.github.com/keijiro/1681052)
A few tweaks had to be made to make this work, mostly just changing the high contrast system and moving the 1/4 grid under the waveform down a little bit because the rendering order had to be changed to match the new track base, which had to match the lasers.
I hope I committed all the correct files, Unity was annoying and modified lots of files as I was poking around so I tried to not commit anything I didn't mean to edit, but I could have missed something I needed.

### DIFF
--- a/Assets/Locales/Options_en-OWO.asset
+++ b/Assets/Locales/Options_en-OWO.asset
@@ -799,69 +799,86 @@ MonoBehaviour:
     m_Localized: Improve performance by disabling the outline on walls.
     m_Metadata:
       m_Items: []
+  - m_Id: 3301721636864
+    m_Localized: Chwanges the twanspawency of the gwid backgwound
+    m_Metadata:
+      m_Items: []
+  - m_Id: 4978382077952
+    m_Localized: '{TextValue}'
+    m_Metadata:
+      m_Items:
+      - id: 16
+  - m_Id: 23812497342464
+    m_Localized: Gwid Twanspawency
+    m_Metadata:
+      m_Items: []
   references:
     version: 1
     00000000:
       type: {class: SmartFormatTag, ns: UnityEngine.Localization.Metadata, asm: Unity.Localization}
       data:
-        m_Entries: 01000000
+        m_Entries: 0000b21e87040000
     00000001:
       type: {class: SmartFormatTag, ns: UnityEngine.Localization.Metadata, asm: Unity.Localization}
       data:
-        m_Entries: 42000000
+        m_Entries: 
     00000002:
       type: {class: SmartFormatTag, ns: UnityEngine.Localization.Metadata, asm: Unity.Localization}
       data:
-        m_Entries: 49000000
+        m_Entries: 
     00000003:
       type: {class: SmartFormatTag, ns: UnityEngine.Localization.Metadata, asm: Unity.Localization}
       data:
-        m_Entries: 83000000
+        m_Entries: 
     00000004:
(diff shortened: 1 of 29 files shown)


Assets/Locales/Options_en-OWO.asset
```diff
@@ -799,69 +799,86 @@ MonoBehaviour:
     m_Localized: Improve performance by disabling the outline on walls.
     m_Metadata:
       m_Items: []
+  - m_Id: 3301721636864
+    m_Localized: Chwanges the twanspawency of the gwid backgwound
```
Top-Cat, line 803: As I've said on other PRs:

> As helpful as this is the way we have crowdin set up this won't get uploaded and will have to be entered there manually anyway

Assets/__Scenes/04_Options.unity
```diff
@@ -16541,6 +16669,440 @@ RectTransform:
     type: 3}
   m_PrefabInstance: {fileID: 820955159}
   m_PrefabAsset: {fileID: 0}
+--- !u!1001 &835319594
```
Top-Cat, line 16672: I know it's not obvious but this object needs adding to the SearchableSection in its parent. Probably should add an option to that class that allows it to populate at runtime but that's outside the scope of this.

Assets/__Scripts/MapEditor/Grid/GridRenderingController.cs
```diff
@@ -78,13 +80,21 @@ private void GridMeasureSnappingChanged(int snapping)
...
-        highContrastBeatPropertyBlock.SetFloat(MainAlpha, Settings.Instance.HighContrastGrids ? 0 : MainAlphaDefault);
-        foreach (Renderer g in gridsToDisableForHighContrast) g.SetPropertyBlock(highContrastBeatPropertyBlock);
+        float gridAlpha = Settings.Instance.GridTransparency;
+        Color newColor = Settings.Instance.HighContrastGrids ? MainColorHighContrast : MainColorDefault;
+        newColor.a = 1f - gridAlpha;
+        Shader shaderToUse =  Shader.Find(newColor.a == 1f ? "Unlit/Color" : "Unlit/Transparent Colored");
+        beatColorPropertyBlock.SetColor(MainColor, newColor);
+        foreach (Renderer g in gridsToColor)
+        {
+            g.material.shader = shaderToUse;
```
Caeden117, line 95: accessing `g.material` will automatically create a new material instance and essentially defeat the whole point of using MaterialPropertyBlocks (saving render calls by re-using the *exact* same material)

I'd recommend finding a different approach that doesnt involve shader switching.

SuperSmay: Would creating two base objects and showing/hiding them depending on what shader is needed be too jank?

Caeden117: not the jankiest thing i've ever done

SuperSmay:
Ok the shader replacement has been changed to hide/unhide different grid base objects instead of accessing `g.material`, and the new setting was added to the `SearchableSection` in the setting's parent as top cat said

SuperSmay:
Ok three commit amends later and it should be all good

PR #318, opened 2021-07-29
add audio latency compensation setting

GalaxyMaster2:
Adds a setting that compensates for audio latency by offsetting all visuals during playback.
AudioTimeSyncController keeps CurrentSeconds (and CurrentBeat) offset, and has new properties CurrentSongSeconds and CurrentSongBeats to expose precise audio timing, which are used by hitsounds and the metronome.

### DIFF
--- a/Assets/__Scripts/MapEditor/AudioTimeSyncController.cs
+++ b/Assets/__Scripts/MapEditor/AudioTimeSyncController.cs
@@ -204,8 +204,8 @@ public class AudioTimeSyncController : MonoBehaviour, CMInput.IPlaybackActions, 
     {
         stopScheduled = true;
         yield return new WaitForSeconds(delaySeconds);
-        if (IsPlaying) TogglePlaying();
         stopScheduled = false;
+        if (IsPlaying) TogglePlaying();
     }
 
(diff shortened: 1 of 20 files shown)


Assets/__Scripts/MapEditor/AudioTimeSyncController.cs
```diff
@@ -175,19 +200,38 @@ private void OnDestroy()
         CurrentSeconds = offsetMS;
     }
 
+    public IEnumerator StopPlayingDelayed(float delaySeconds)
+    {
+        stopScheduled = true;
+        yield return new WaitForSeconds(delaySeconds);
+        if (IsPlaying) TogglePlaying();
+        stopScheduled = false;
```
Top-Cat, line 208: Won't this always be set by TogglePlaying?

GalaxyMaster2: Currently, yes.
But if it were to be called when not playing it could result in stopScheduled staying true.
That being said I could probably move it up a line to prevent StopCoroutine being called unnecessarily.

PR #326, opened 2021-08-10
enable various forms of timeline input during playback

GalaxyMaster2:
Enables:
- clicking and dragging on timeline
- clicking on bookmarks
- bookmark navigation keys

during playback.

This is done by pausing and resuming playback on pointer down/up. I have not found any obvious cases of undesired behavior with this approach, but if you have concerns feel free to bring them up.


Caeden117 (approved):
k

PR #327, opened 2021-08-11
remove old song offset code

GalaxyMaster2:
It was horribly broken and now that songTimeOffset is handled by adjusting the audio data it's interfering and no longer needed.

Also includes a bonus fix for the song offset warning


Caeden117 (approved):
k

PR #328, opened 2021-08-12
fix accumulating floating point errors on track z-position

GalaxyMaster2:
Fixes an elusive bug that's been around for a while where under some conditions the note track could become persistently slightly offset causing notes to become transparent when they shouldn't.


Caeden117 (approved):
k

PR #336, opened 2021-09-06
Alpha visualization

SuperSmay:
Adds the option of visualizing Chroma event alphas using event height. Also adds a new flat pyramid event model.


Assets/__Scripts/Map/Events/EventAppearanceSO.cs
```diff
@@ -28,6 +29,20 @@ public class EventAppearanceSO : ScriptableObject
     [SerializeField] private Vector3 pyramidFadeShaderOffset = Vector3.zero;
     [SerializeField] private float pyramidDefaultFadeSize = 50f;
     [SerializeField] private float pyramidBoostEventFadeSize = 10f;
+    [Header("FlatPyramid")]
```
Top-Cat, line 32: Why add code just to be commented?
Delete code that you don't need

Assets/__Scripts/Map/Events/EventAppearanceSO.cs
```diff
@@ -132,27 +147,34 @@ public class EventAppearanceSO : ScriptableObject
...
-        e.ChangeFadeSize(e.UsePyramidModel ? pyramidDefaultFadeSize : cubeDefaultFadeSize, false);
+        e.ChangeFadeSize(defaultFadeSize[e.EventModel], false);
 
         if (Settings.Instance.VisualizeChromaGradients) e.UpdateGradientRendering();
 
         e.UpdateMaterials();
     }
 }
+
+public enum EventModelType
```
Top-Cat, line 175: Unless I'm missing something this is the most wasted enum. You use int everywhere and only cast from this enum once.
(You could be using it in a scene somewhere I haven't checked)
But you don't even use it in UpdateOldSettings???

SuperSmay:
I changed where the properties for the event material for each model are stored, let me know if this seems like a decent way to do it. I like it more because it's much easier to figure out what properties go to what and doesn't add a large number of fields to `EventAppearenceSO`, but I could see them being stored and accessed this way not being organizationally ideal, because they aren't in `EventAppearenceSO` anymore.

PR #337, opened 2021-09-06
Bookmark colors

SuperSmay:
Adds the ability for bookmarks to have a color stored with them, and adds a color selections dialog box to change the colors of said bookmarks

### DIFF
--- a/Assets/__Scripts/HSVPicker/UI/ColorPicker.cs
+++ b/Assets/__Scripts/HSVPicker/UI/ColorPicker.cs
@@ -33,7 +33,6 @@ public class ColorPicker : MonoBehaviour
         {
             if (CurrentColor == value)
                 return;
-
             _red = value.r;
             _green = value.g;
             _blue = value.b;
--- a/Assets/__Scripts/HSVPicker/UI/ColorPicker.cs
+++ b/Assets/__Scripts/HSVPicker/UI/ColorPicker.cs
@@ -33,6 +33,7 @@ public class ColorPicker : MonoBehaviour
         {
             if (CurrentColor == value)
                 return;
+
             _red = value.r;
             _green = value.g;
             _blue = value.b;
--- a/Assets/__Scripts/Map/Bookmarks/BeatmapBookmark.cs
+++ b/Assets/__Scripts/Map/Bookmarks/BeatmapBookmark.cs
@@ -12,10 +12,10 @@ public class BeatmapBookmark : BeatmapObject
         _time = RetrieveRequiredNode(node, "_time").AsFloat;
         _name = RetrieveRequiredNode(node, "_name");
         if (node.HasKey("_color")) _color = ColorFromJSONArray(RetrieveRequiredNode(node, "_color").AsArray);
-        else _color = GetNewColor();
+        else _color = GetRandomColor();
     }
 
-    private Color GetNewColor()
+    private Color GetRandomColor()
     {
         Color color = UnityEngine.Random.ColorHSV(0, 1, 0.75f, 0.75f, 1, 1);
         return color;
@@ -35,11 +35,24 @@ public class BeatmapBookmark : BeatmapObject
         return colorArray;
     }
 
-    public BeatmapBookmark(float time, string name, Color color)
+    public BeatmapBookmark(float time, string name)
     {
         _time = time;
         _name = name;
-        _color = color;
+        _color = GetRandomColor();
+        //_color = color;
+    }
+
+    public void NewColor()
+    {
+        PersistentUI.Instance.ShowColorInputBox("Mapper", "bookmark.dialog", HandleNewBookmarkColor, "bookmark.dialog.default");
+    }
+
+    private void HandleNewBookmarkColor(object res)
+    {
+        if (res == null || res.GetType() != typeof(Color)) return;
+        _color = (Color)res;
+
     }
 
     public override JSONNode ConvertToJSON()
--- a/Assets/__Scripts/Map/Bookmarks/BeatmapBookmark.cs
+++ b/Assets/__Scripts/Map/Bookmarks/BeatmapBookmark.cs
@@ -11,23 +11,10 @@ public class BeatmapBookmark : BeatmapObject
     {
         _time = RetrieveRequiredNode(node, "_time").AsFloat;
         _name = RetrieveRequiredNode(node, "_name");
-        if (node.HasKey("_color")) _color = ColorFromJSONArray(RetrieveRequiredNode(node, "_color").AsArray);
+        if (node.HasKey("_color")) _color = RetrieveRequiredNode(node, "_color");
         else _color = UnityEngine.Random.ColorHSV(0, 1, 0.75f, 0.75f, 1, 1);
     }
 
-    private Color ColorFromJSONArray(JSONArray color)
-    {
-        return new Color(color[0], color[1], color[2]);
-    }   
-
-    private JSONArray ColorToJSONArray(Color color)
-    {
-        JSONArray colorArray = new JSONArray();
-        colorArray[0] = color.r;
-        colorArray[1] = color.g;
-        colorArray[2] = color.b;
-        return colorArray;
-    }
 
     public BeatmapBookmark(float time, string name)
     {
@@ -42,7 +29,7 @@ public class BeatmapBookmark : BeatmapObject
         JSONNode node = new JSONObject();
         node["_time"] = Math.Round(_time, decimalPrecision);
         node["_name"] = _name;
-        node["_color"] = ColorToJSONArray(_color);
+        node["_color"] = _color;
         return node;
     }
 
--- a/Assets/__Scripts/Map/Bookmarks/BeatmapBookmark.cs
+++ b/Assets/__Scripts/Map/Bookmarks/BeatmapBookmark.cs
@@ -11,13 +11,23 @@ public class BeatmapBookmark : BeatmapObject
     {
         _time = RetrieveRequiredNode(node, "_time").AsFloat;
         _name = RetrieveRequiredNode(node, "_name");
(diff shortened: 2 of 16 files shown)


Assets/__Scripts/Map/Bookmarks/BeatmapBookmark.cs
```diff
@@ -1,25 +1,48 @@
...
 {
     public BeatmapBookmark(JSONNode node)
     {
         _time = RetrieveRequiredNode(node, "_time").AsFloat;
         _name = RetrieveRequiredNode(node, "_name");
+        if (node.HasKey("_color")) _color = ColorFromJSONArray(RetrieveRequiredNode(node, "_color").AsArray);
+        else _color = UnityEngine.Random.ColorHSV(0, 1, 0.75f, 0.75f, 1, 1);
+    }
+
+    private Color ColorFromJSONArray(JSONArray color)
```
Top-Cat, line 18: There are already json helpers for reading and writing colours

Assets/__Scripts/MapEditor/UI/Bookmarks/BookmarkContainer.cs
```diff
@@ -44,7 +46,8 @@ public void OnPointerClick(PointerEventData eventData)
                 PersistentUI.Instance.ShowDialogBox("Mapper", "bookmark.delete", HandleDeleteBookmark, PersistentUI.DialogBoxPresetType.YesNo);
                 break;
             case PointerEventData.InputButton.Right:
-                PersistentUI.Instance.ShowInputBox("Mapper", "bookmark.update.dialog", HandleNewBookmarkName, null, data._name);
+                if (Input.GetKey(KeyCode.LeftShift)) PersistentUI.Instance.ShowColorInputBox("Mapper", "bookmark.update.color", HandleNewBookmarkColor, GetComponent<Image>().color);
```
Top-Cat, line 49: We don't use Input.GetKey here 🤨
If you're not going to make this individually rebindable at least use `CMInput.IUtilsActions` that supplies binds for Ctrl, Shift and Alt

Assets/__Scripts/UI/CM_ColorInputBox.cs
```diff
@@ -0,0 +1,90 @@
...
+using UnityEngine.InputSystem;
+using System.Linq;
+using System.Collections.Generic;
+
+public class CM_ColorInputBox : MenuBase
+{
+    [SerializeField] private ColorPicker ColorInputField;
+    [SerializeField] private TextMeshProUGUI UIMessage;
+    [SerializeField] private CanvasGroup group;
+    private Action<object> resultAction;
```
Top-Cat, line 15: Why is this object and not Color?

Assets/__Scripts/MapEditor/UI/Bookmarks/BookmarkManager.cs
```diff
@@ -51,7 +51,6 @@ internal void HandleNewBookmarkName(string res)
         container.name = newBookmark._name;
         container.Init(this, newBookmark);
         container.RefreshPosition(timelineCanvas.sizeDelta.x + CANVAS_WIDTH_OFFSET);
-
```
Top-Cat, line 54: y u decrease readability?

Assets/__Scripts/MapEditor/UI/Bookmarks/BookmarkContainer.cs
SuperSmay: How would I use `Cminput.IUtilsActions`? I can't seem to figure out how it works and I don't know anything about input systems. I tried looking at other places it's used but it doesn't make sense to me.

Assets/__Scripts/UI/CM_ColorInputBox.cs
SuperSmay: This was because I was following what the other dialog boxes do. They invoke `result` with `null` when cancelled, and Color is a non-nullable type, so invoking with null wouldn't work. If there is no reason for the input box to invoke `result` when it is cancelled, then it can be switched to Color and `result` can just be ignored.

Caeden117: Would using [`Color?`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types) work?

Top-Cat: It should since we updated unity

SuperSmay:
Everything should now be fixed except for the input binding

Assets/__Scripts/MapEditor/UI/Bookmarks/BookmarkContainer.cs
Top-Cat: You just implement the interface

Caeden117: Unfortunately, due to the fact that bookmarks are created/removed dynamically, it wouldn't immediately work with the [`CMInputCallbackInstaller`](https://github.com/Caeden117/ChroMapper/blob/dev/Assets/__Scripts/Input/ChroMapper/CMInputCallbackInstaller.cs). This class is what allows the Input System to be as easy as "just implement the interface", but currently only works on MonoBehaviors that exist from the beginning.

To get the input interface to bind here, we'd either need to move this bookmark functionality out to the `BookmarkManager` which *would* be picked up by the callback installer, OR we would have to publicly expose the `FindAndInstallCallbacksRecursive` method (or create a similar method for one singular monobehaviour/transform) and dynamically install callbacks for each `BookmarkContainer`.

Personally I'd go for the former since it's less work to do, and have the `OnPointerClick` code simply forward to the `BookmarkManager`.

SuperSmay: Is this how the interface would be implemented? 
```cs
public class BookmarkManager : MonoBehaviour, CMInput.IBookmarksActions, CMInput.IUtilsActions
...
    public void OnControlModifier(InputAction.CallbackContext context)
        {
        }

    public void OnAltModifier(InputAction.CallbackContext context)
        {
        }

    public void OnShiftModifier(InputAction.CallbackContext context)
        {
            //Do things
        }

    public void OnMouseMovement(InputAction.CallbackContext context)
        {
        }
```
I tried doing that with a `Debug.Log` to see if anything was happening when I pressed the shift key and nothing was logged, so I think I did it wrong but this was the best I could figure out from looking at other places this stuff was used

Caeden117: That *should* work, make sure you didnt have `Log`s suppressed in the Editor.

SuperSmay: I closed Unity and came back to it later and it works now. Not a clue as to why. Either way thanks for the help and info and I'll see if I can do something useful with this

SuperSmay:
The color modifier now uses a remappable binding implemented in `BookmarkManager`.

Assets/__Scripts/Map/Bookmarks/BeatmapBookmark.cs
Top-Cat: This is not fixed, you're still using this helper for no reason.

Assets/__Scripts/HSVPicker/UI/ColorPicker.cs
```diff
@@ -33,7 +33,6 @@ public Color CurrentColor
         {
             if (CurrentColor == value)
                 return;
-
```
Top-Cat, line 36: Again, why?

Assets/__Scripts/Map/Bookmarks/BeatmapBookmark.cs
```diff
@@ -1,25 +1,33 @@
...
+        else _color = UnityEngine.Random.ColorHSV(0, 1, 0.75f, 0.75f, 1, 1);
     }
 
+
     public BeatmapBookmark(float time, string name)
     {
         _time = time;
         _name = name;
+        _color = UnityEngine.Random.ColorHSV(0, 1, 0.75f, 0.75f, 1, 1);
+        //_color = color;
```
Top-Cat, line 22: ?

Assets/__Scripts/MapEditor/UI/Bookmarks/BookmarkContainer.cs
```diff
@@ -13,7 +13,8 @@ public void Init(BookmarkManager manager, BeatmapBookmark data)
         if (this.data != null) return;
         this.data = data;
         this.manager = manager;
-        GetComponent<Image>().color = Random.ColorHSV(0, 1, 0.75f, 0.75f, 1, 1);
+        GetComponent<Image>().color = data._color;
+            //Random.ColorHSV(0, 1, 0.75f, 0.75f, 1, 1);
```
Top-Cat, line 17: ?

SuperSmay:
Ok I think I fixed all the comments I forgot to remove, hopefully I didn't miss any

Assets/__Scripts/MapEditor/UI/Bookmarks/BookmarkContainer.cs
```diff
@@ -66,6 +68,24 @@ public void OnPointerUp(PointerEventData eventData)
...
+
+    public void OnShiftModifier(InputAction.CallbackContext context)
+    {
+        Debug.LogError(context);
+        
+    }
+
+    public void OnMouseMovement(InputAction.CallbackContext context)
+    {
+    }
```
Caeden117, line 87: Do you mind removing the now unused code since `BookmarkManager` handles the input stuff now?

SuperSmay: Knew I was forgetting something. Gone now.

PR #339, opened 2021-09-11
Rewrite dev console

Top-Cat:
Completely replaced the built-in unity console with a custom UI element.

When an exception occurs a flag is shown, clicking on it generates a report with the user's system specs, the exception message and the previous 20 log lines. Example: https://paste.ee/p/CDCcM

There are commented out methods to write this to a local file or send it to pastebin instead (which has an api limit of 25 per day that I quickly used up)

As this now also shows normal log messages it makes scripting easier as you no longer need to go and find your log files to see console.log output.

This also starts its own file writer where it writes everything it catches. Some log messages are still sent to the normal Player.log such as the asset GC and other weird unity stuff that I guess isn't using the proper APIs.

image


Assets/__Scripts/UI/LogLineUI.cs
```diff
@@ -0,0 +1,149 @@
...
+               "RAM: " + SystemInfo.systemMemorySize + " MB\n" +
+               "OS: " + SystemInfo.operatingSystem;
+    }
+
+    private string Heading(string text, bool first = false)
+    {
+        return (first ? "" : "\n\n\n") + $"{Seperator}\n{text}\n{Seperator}\n";
+    }
+
+    public IEnumerator SendToPastebin()
```
Caeden117, line 60: Honestly I'd rename this to something like `GenerateBugReport`, especially when Pastebin isn't being used at the moment.

Assets/__Scripts/UI/LogLineUI.cs
```diff
@@ -0,0 +1,149 @@
...
+                Application.OpenURL(result);
+            }
+        }
+    }
+    
+    private static IEnumerator CreateAsync(string text, string title = "Untitled", string language = "csharp", int visibility = 1, string expiration = "N")
+    {
+        //yield return WriteErrorToFile(text);
+        yield return UploadToPasteee(text, title);
+        //yield return UploadToPastebin(text, title);
```
Caeden117, line 147: Would be pretty gamer if we had a dialog box that allowed the user to pick between the available options (although if this happens I dont think the coroutine would be necessary)

Assets/__Scripts/UI/DevConsole.cs
```diff
@@ -0,0 +1,198 @@
...
+    private string GetLogFolder()
+    {
+#if UNITY_STANDALONE_WIN
+        return Path.Combine(Environment.GetEnvironmentVariable("AppData"), "..", "LocalLow", Application.companyName, Application.productName);
+#elif UNITY_STANDALONE_OSX
+        return "~/Library/Logs/Unity";
+#elif UNITY_STANDALONE_LINUX
+        return Path.Combine("~/.config/unity3d", Application.companyName, Application.productName);
+#endif
+    }
```
Caeden117, line 163: Couldn't this all be simplified to a simple call to [`Application.persistentDataPath`](https://docs.unity3d.com/ScriptReference/Application-persistentDataPath.html)?

Assets/__Scripts/UI/LogLineUI.cs
Top-Cat: Guess I removed the comment but I moved from pastebin to pasteee because pastebin both auto-deleted my initial tests and limits you to 25 uploads per day (I assume per ip and not per key) but I left the code here "just in case".

I could allow you to pick a file instead but I'm not sure there's really a downside to always uploading it.

Assets/__Scripts/UI/LogLineUI.cs
Top-Cat: This may be a derp moment

Assets/__Scripts/UI/DevConsole.cs
Top-Cat: This may be a derp moment

PR #341, opened 2021-09-19
Fix bookmark color error on beatmap save

SuperSmay:
Also might fix that random color not being random issue because unity's random color isn't used anymore


Caeden117 (approved):
k

PR #342, opened 2021-09-20
Fix NullReferenceException when searching options

Top-Cat:


Caeden117 (approved):
this is a monumental PR! such an elegant solution to a difficult problem!

k

PR #359, opened 2021-11-27
improvements to the "playing" ui mode (and some other things)

GalaxyMaster2:
- Fixes the crouch keybind in playing mode (defaults to S)
- Saves and restores the camera position when going in and out of playing mode
- Fixes the spawn offset calculation (and moved HJD calculation code to helper class)
- Shows the song timeline when paused (in playing and preview modes)
- UI fade transition times reduced to 0.2 seconds (from 1 second)
- Disables grid line on the surface of notes when the placement grid is not visible
- Obstacle fade distance during playback now follows the spawn offset for notes
- Camera z-position tweaked in playing mode (from -7 to -6)

### DIFF
--- a/Assets/__Scripts/Map/SpawnParameterHelper.cs
+++ b/Assets/__Scripts/Map/SpawnParameterHelper.cs
@@ -1,6 +1,4 @@
-﻿using UnityEngine;
-
-public class SpawnParameterHelper : MonoBehaviour
+﻿public static class SpawnParameterHelper
 {
     public static float CalculateHalfJumpDuration(float noteJumpSpeed, float startBeatOffset, float bpm) {
         var halfJumpDuration = 4f;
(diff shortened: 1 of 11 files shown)


Assets/__Scripts/Map/SpawnParameterHelper.cs
```diff
@@ -0,0 +1,24 @@
+﻿using UnityEngine;
+
+public class SpawnParameterHelper : MonoBehaviour
```
Caeden117, line 3: Considering this is all static methods, this class really does not need the extra baggage of `MonoBehaviour`. It's probably more fitting to have this be a `public static class`.

GalaxyMaster2: makes sense

GalaxyMaster2:
Made 2 more changes:
- Shortened the transition times for the ui mode selector
- Hide BPM changes in playing/preview modes

GalaxyMaster2:
Another one:
- Restore cursor position after unlocking cursor instead of resetting to center

PR #361, opened 2021-11-29
trackpad scrolling improvements

GalaxyMaster2:
Submitting this as a draft for now. This is an initial implementation of improved trackpad scrolling. It does this by accounting for the size of the scroll value and keeping track with an internal counter. This changes trackpad scrolling from uncontrollably fast to natural and adaptive.

To do:
- [ ] Verify the correct `scrollSizeDivisor` on macOS (will need outside help for this)
- [ ] Verify that scrolling with regular mouse wheels is not impacted in any way
- [ ] Add similar logic to scrolling the cursor precision (Ctrl+Scroll)
- [ ] (Potentially) add a settings toggle
- [ ] (Potentially) add a scroll speed scalar setting
- [ ] (Potentially) add some visualization for the internal scroll step counter

### DIFF
--- a/Assets/__Scripts/MapEditor/AudioTimeSyncController.cs
+++ b/Assets/__Scripts/MapEditor/AudioTimeSyncController.cs
@@ -220,14 +220,17 @@ public class AudioTimeSyncController : MonoBehaviour, CMInput.IPlaybackActions, 
 
                 var scrollSize = value / scrollSizeDivisor;
                 internalScrollStepCounter += scrollSize;
-                if (Mathf.Abs(internalScrollStepCounter) >= 0.5f)
+
+                // +1 beat if we're going forward, -1 beat if we're going backwards
+                var direction = Mathf.Sign(internalScrollStepCounter);
+                var beatShiftRaw = 0f;
+                while (Mathf.Abs(internalScrollStepCounter) >= 0.5f)
                 {
-                    // +1 beat if we're going forward, -1 beat if we're going backwards
-                    var direction = value > 0 ? 1f : -1f;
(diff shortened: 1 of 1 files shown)


GalaxyMaster2:
There seems to be a bug in unity on linux where trackpad scrolling specifically is different in the editor compared to standalone builds. In standalone builds the scroll values are reported as integer values, making the scrolling too fast and less precise. In the editor the scrolling works similar to on windows, with float scroll values and normal speed.

Upgrading unity to 2022.1.0a16 changed this, but instead of fixing the broken scrolling in standalone builds, it simply broke the editor in the same way. At least it's consistent I guess.

This problem makes a scroll speed multiplier setting pretty much necessary for proper trackpad scrolling on linux.

yozzozo:
Suggestion (if you haven't considered it already) -- if you're making a settings toggle for the scroll speed multiplier, could there be an additional one for "fast" vs normal speed?  For example, trackpad scrolling would normally be at rate X, and holding down shift (or whatever you set the modifier to) and scrolling would be at rate Y.

GalaxyMaster2:
> Suggestion (if you haven't considered it already) -- if you're making a settings toggle for the scroll speed multiplier, could there be an additional one for "fast" vs normal speed? For example, trackpad scrolling would normally be at rate X, and holding down shift (or whatever you set the modifier to) and scrolling would be at rate Y.

The setting I mentioned would just be a slider or a number input, not a fixed extra multiplier to be toggled on or off. As for your suggestion of a keybind to shift to a 2nd scroll rate, I don't personally think there will be much of a need for it, but I may consider it in the future.

PR #366, opened 2021-12-14
fix difficulty label not showing immediately on copy

GalaxyMaster2:
This had been bugging me for a while and it turned out to be a one line fix.

### DIFF
--- a/Assets/__Scripts/UI/SongEditMenu/DifficultySelect.cs
+++ b/Assets/__Scripts/UI/SongEditMenu/DifficultySelect.cs
@@ -350,6 +350,7 @@ public class DifficultySelect : MonoBehaviour
 
             diffs[row.Name] = new DifficultySettings(map, true);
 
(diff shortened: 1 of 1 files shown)


Top-Cat:
I think my original assumption is that you wouldn't actually want to copy a custom diff name so I would have cleared the custom data value. (It's possible originally we didn't copy it but since adding copying for colours it's now being copied too?)

GalaxyMaster2:
> I think my original assumption is that you wouldn't actually want to copy a custom diff name so I would have cleared the custom data value. (It's possible originally we didn't copy it but since adding copying for colours it's now being copied too?)

To me it makes sense to also copy the custom label, but I could see arguments for either way. Shouldn't be hard to just not copy the label instead, but either is preferable to the current broken behavior.

Top-Cat:
I'd like a second opinion, I do agree the current behaviour needs changing

GalaxyMaster2:
Seems good to me, done

PR #373, opened 2022-02-07
add render scale setting

GalaxyMaster2:
Useful setting for performance or visual quality improvements. Setting scales the total number of pixels rendered, and is implemented as an int percentage, ranging from 1% to 400% (which are the limits set by URP). From testing it seems that values between ~90% - 110% are treated as 100% by the renderer, likely to avoid unnecessary blurring.


Caeden117 (approved):
gamering

PR #374, opened 2022-02-09
update max jump duration in HJD calculation

GalaxyMaster2:
In a recent game update this was changed from 36 to 35.998, presumably to combat the 36 JD bug


Top-Cat:
Now the 35.998 JD bug

PR #375, opened 2022-02-11
fix floating point errors introduced to atsc.CurrentBeat when changing editor scale

GalaxyMaster2:
This fixes an annoying bug I've been chasing for a while where for seemingly no reason I was placing notes slightly off-grid.
EditorScaleController calls `atsc.MoveToTimeInSeconds`, changing this to `MoveToTimeInBeats` avoids introducing floating point errors.


Caeden117 (approved):
gaming

PR #377, opened 2022-02-19
Don't yeet lightId from env enhancements

XAce1337manX:
Apparently `_lightID` and `_lightId` both work, so using `_lightId` because that's what documentation says.


Top-Cat:
I can't see any code in Chroma that supports the assertion that both work and while the documentation may have `_lightId` the same constant is used as for the map event that specifies `_lightID` https://github.com/Aeroluna/Heck/blob/master/Chroma/ChromaController.cs#L34

Therefore I believe this PR would not actually fix lightId in env enhancements. Stop handling the `_lightId` key and write it in the correct case.

PR #381, opened 2022-03-16
Add barebone support for BeatmapV3

CCDV2:
#### Add barebone support for BeatmapV3, including
- V3 ColorNotes & BombNotes
- V3 Obstacles
- BPMChanges
- ColorBoost   
Above are treated as compatible classes, and resuing the old logic. (Basically they are all what I usually use for mapping)
- Sliders
- Chains(BurstSliders)   
Above are newly introduced in V3, only available in V3 mode   
- Conversion between v2 and v3   
Basic v2 features(no mod needed features) could be converted. When v3 converting to v2, slider and chain will get lost.   
> The basic idea of new implementation is trying not to be obtrusive and reusing the old logic(like notes, obstacles and events). When new logic has to be added to exsiting code, it will firstly check `Settings.Instance.Load_MapV3` to ensure editor isn't editing a v2 map. `Settings.Instance.Load_MapV3` is only  modified when entering a map or switching the map version.

#### Exisitng problems
- Visual effect for chain is still not perfect, and its detection for head note is not perfect.
- Counter may not count on a no-head chain.
- Keybind for slider `T` conflict with track.   
- Slider & Chain are using chunk loading now, resulting them spawn not in the same distance as note.
For details, see `V3CONTROL.md`

#### Could be introduced features in future
- Angle offest for notes
- Compatibility with mod
- UI for slider & chain spawn(instead of key only)
- Make chain & slider shown on the top-left box(but how?)
- New environment & new event type

#### Newly imported packages
- [SplineMesh](https://assetstore.unity.com/packages/tools/modeling/splinemesh-104989)
- [Bezier Solution](https://assetstore.unity.com/packages/tools/level-design/bezier-solution-113074) (imported at begining, later removed)

#### Bug introduced by this PR
I'm sorry for a huge PR. If there is any bug only happens after this PR, I could help check.


JD-The-65th:
BPM changes are currently broken, and apply an unintended affect while playing a map in game.

Assets/__Scripts/Map/BeatSaberMapV3.cs
```diff
@@ -0,0 +1,329 @@
...
+            var lightColorEventBoxGroups = new JSONArray();
+            foreach (var l in LightColorEventBoxGroups) lightColorEventBoxGroups.Add(l);
+
+            var lightRotationEventBoxGroups = new JSONArray();
+            foreach (var l in lightRotationEventBoxGroups) lightRotationEventBoxGroups.Add(l);
+
+            var basicEventTypesWithKeywords = new JSONArray();
+            foreach (var b in BasicEventTypesWithKeywords) basicEventTypesWithKeywords.Add(b);
+
+            MainNode["bpmEvents"] = CleanupArray(bpmEvents, "b");
```
Caeden117, line 110: Let's not use Official BPM Changes. They have been infamously buggy, and from @JD-The-65th's comment, likely still are to this day. For the time being, let's continue to put them in Custom Data.

I asked Aeroluna a few days ago about how Custom Data will be stored in Beatmap v3, and he told me this:
- Custom data in `Info.dat` will still be stored in `_customData`
- Custom data in the root of the Difficulty (like our BPM Changes) will be stored in `customData`
- Custom data per object (Chroma event data, NE data, etc.) will be stored in `cd`

JD-The-65th: Also, if you want to experience the effects of official BPM changes yourself, look no further than lawless Africa https://beatsaver.com/maps/1d233

CCDV2: Thanks for the suggestion. Then I think I could leave bpm change still for customdata to parse. I don't have to take care of it(However I'm not sure what effect it will be with slider and chain).
Besides, for customdata, I think I need to wait for a chromma supprot and test its functionality,

CCDV2:
#### Update on new 3 commits
- Revert `BpmChange` to original. 
- When a map is v3, it reads from and saves to `customData`
- When a `BeatmapObject` is v3, it reads from and saves to `cd`  
I playtested [Pure-Hearted Armeria](https://bsaber.com/songs/21b0b/) converted to v3. It works fine to me(although I cannot test chroma at the moment).

JD-The-65th (commented):
I suggest changing the keybinds for the Chains and Sliders to C and I respectfully (C for Chains, I for Sliders). Not required, just makes a bit more sense

CCDV2:
#### Update on new 4 commits 
I was making a v3 map and found some issue. Therefore I revised some places.
- Spline height seems to be a constant in-game, with no relation to start-end distance. Therefore I fix height to 3. It now seems similar.
- Chain now looks good when the center degree between head & tail is < 180(those two points are on the same circle). It should have the logic when rotation anlge > 180, but I haven't figured out what the pattern looks like in-game, so currently just ignore it.

#### Found issue
When slider is used for decoration(like the image), it will have a noticeable stuck while loading this slider chunk. Because all of the sliders have to recompute its spline.
Webp net-resizeimage

Caeden117:
> Spline height seems to be a constant in-game, with no relation to start-end distance. Therefore I fix height to 3. It now seems similar.

From my understanding, arcs also have a clamped floor and left/right sides, depending on the environment (ex. Linkin Park is a much more enclosed environment than something like Billie). At some point, we should make these bounds platform-specific.

Assets/__Scripts/UI/SongInfoEditUI.cs
```diff
@@ -542,6 +542,8 @@ private void HandleEditMapButtonPressed(int r)
                     .BeatmapCharacteristicName;
                 Settings.Instance.LastLoadedDiff = BeatSaberSongContainer.Instance.DifficultyData.Difficulty;
                 BeatSaberSongContainer.Instance.Map = map;
+                Settings.Instance.Load_MapV3 = map.Version[0] == '3';
+                if (Settings.Instance.Load_MapV3) Settings.Instance.GridTransparency = 0.75f;
```
Caeden117, line 546: No.

`Settings.Instance` will always be loaded from, and saved to, a file on disk. You are effectively changing a setting without user consent. I would much rather leave the `GridTransparency` setting untouched and have the user change it themselves if they have vision concerns.

Caeden117 (changes requested):

GalaxyMaster2:
> > Spline height seems to be a constant in-game, with no relation to start-end distance. Therefore I fix height to 3. It now seems similar.
> 
> From my understanding, arcs also have a clamped floor and left/right sides, depending on the environment (ex. Linkin Park is a much more enclosed environment than something like Billie). At some point, we should make these bounds platform-specific.

I don't think that's what that is about, but I could be wrong. I think environment clamping isn't implemented at all currently.

GalaxyMaster2:
Having used this for a bit I have some feedback:

Arcs:
- Their shape seems quite close to how it is in game, so good job on that.
- If one of the notes is a dot note, making an arc does not work. If you manually edit the cut direction to 8, it is treated as down. In game the effect seems to be the same as setting the multiplier to 0 (I have not fully verified this however).
- Mid anchor mode is ignored. I haven't quite figured out how it works in game but I do know that it only applies in some situations, particularly when the notes are in line with their cut direction (so the arc would end up with no rotation).
- Performance seems to be quite bad. With only a couple arcs present I'm getting stutters when scrolling fast and slight hiccups when pressing play. This might also apply to chains, I haven't tested many of those yet.

Chains:
- I'm sure you know this already, but the head note looks quite different from in game. Something to improve.
- Slice count on alt+scroll is limited to 3, but in game it works perfectly down to 2 (and even 1, though that is probably outside of intended use).
- In game slices are just as wide as normal notes, but here they are significantly smaller.
- Slices seem to have some unintended rotation on other axes.
- Already known but moving the head note after placing a chain doesn't update the model to a chain head in situations where it should.
- The way chains curve doesn't seem to match perfectly to how it is in game. If I'm not wrong it looks like here it uses a cicrular curve, while in game a bezier curve is used.
- Chains only have a hitsound for the head note. Currently this is accurate to how it is in game, however this is supposedly still being worked on, so something to keep an eye on.

Overall:
- Arcs and chains don't respect the Simple Blocks option. They retain the shiny look.
- I would like to see alt+drag controls for both arcs and chains.
- We might also want to consider more direct ways of placing arcs and chains as an alternative placement option.
- I'm getting NullReferenceExceptions when alt+dragging, even on objects other than arcs/chains.
- I would like at least support for converting 2.6.0 maps with arcs to 3.0.0, even if saving in that format won't be added. I had to manually convert arcs on one of my maps.

GalaxyMaster2:
Another thing, angle offset on notes is ignored right now, this should be implemented.
Possibly consider changing alt+scroll on notes to modify that instead of cut direction, or have a different key combination that does that.

GalaxyMaster2:
When placing a note in a v3 map and inspecting it with node editor, it shows the note data in v2 format. It does seem to get converted when saving and exiting though.

Assets/__Scripts/UI/SongInfoEditUI.cs
CCDV2: One of my concern is that when a user sets Load_MapV3 to true and opens a v2 map, he will ge nothing because it's not the same json format. However if he misclicks to save or is auto-saved then previous notes are all overwritten and gone, which may not be the intention..

CCDV2:
> Having used this for a bit I have some feedback:
> 
> Arcs:
> 
> * Their shape seems quite close to how it is in game, so good job on that.
> * If one of the notes is a dot note, making an arc does not work. If you manually edit the cut direction to 8, it is treated as down. In game the effect seems to be the same as setting the multiplier to 0 (I have not fully verified this however).
> * Mid anchor mode is ignored. I haven't quite figured out how it works in game but I do know that it only applies in some situations, particularly when the notes are in line with their cut direction (so the arc would end up with no rotation).
> * Performance seems to be quite bad. With only a couple arcs present I'm getting stutters when scrolling fast and slight hiccups when pressing play. This might also apply to chains, I haven't tested many of those yet.
> 
> Chains:
> 
> * I'm sure you know this already, but the head note looks quite different from in game. Something to improve.
> * Slice count on alt+scroll is limited to 3, but in game it works perfectly down to 2 (and even 1, though that is probably outside of intended use).
> * In game slices are just as wide as normal notes, but here they are significantly smaller.
> * Slices seem to have some unintended rotation on other axes.
> * Already known but moving the head note after placing a chain doesn't update the model to a chain head in situations where it should.
> * The way chains curve doesn't seem to match perfectly to how it is in game. If I'm not wrong it looks like here it uses a cicrular curve, while in game a bezier curve is used.
> * Chains only have a hitsound for the head note. Currently this is accurate to how it is in game, however this is supposedly still being worked on, so something to keep an eye on.
> 
> Overall:
> 
> * Arcs and chains don't respect the Simple Blocks option. They retain the shiny look.
> * I would like to see alt+drag controls for both arcs and chains.
> * We might also want to consider more direct ways of placing arcs and chains as an alternative placement option.
> * I'm getting NullReferenceExceptions when alt+dragging, even on objects other than arcs/chains.
> * I would like at least support for converting 2.6.0 maps with arcs to 3.0.0, even if saving in that format won't be added. I had to manually convert arcs on one of my maps.

I have to admit there is a lot needed to be done. I will implement them if I want to use them making a map.

CCDV2:
> We might also want to consider more direct ways of placing arcs and chains as an alternative placement option.   

One solution could be like placing an obstalce. But I don't think it gets better enough..   

> I would like to see alt+drag controls for both arcs and chains.   

How to drag a slider? Its indicators are out of grid, which may be hard to be dragged precisely.  
Currently chains and sliders can mirror or cut/paste, may it help to some extent.  


> The way chains curve doesn't seem to match perfectly to how it is in game. If I'm not wrong it looks like here it uses a cicrular curve, while in game a bezier curve is used.   

Chains' curve really confuse me. There could be a lot of weird cases when rotation angle is >= 180.. 
    
For v3 fomat note, you see v2 json because it is converted to v2 after reading from json. If we want to support anlge offset for v3 notes, then I think we need to refractor previous code to make a factory for notes instead of init a v2 note directly. But I'm not sure how many places need to be refractored.

SuperSmay:
> How to drag a slider? Its indicators are out of grid, which may be hard to be dragged precisely. Currently chains and sliders can mirror or cut/paste, may it help to some extent.

Make the end points of arcs (and maybe chains) draggable nodes that can also be rotated, much like normal notes. There would be no need to have the arc indicators be draggable themselves as you can adjust them with the scroll keybinds.

CCDV2:
> Make the end points of arcs (and maybe chains) draggable nodes that can also be rotated, much like normal notes. There would be no need to have the arc indicators be draggable themselves as you can adjust them with the scroll keybinds.

But normally there will be a note at end point. It cannot tell you want to drag the note or drag the slider.

SuperSmay:
> But normally there will be a note at end point. It cannot tell you want to drag the note or drag the slider.

That is true, but if you need to move the end point it seems likely that you either moved the note already or don't have a note there at all. If there is one though you could simply move the note out of the way then put it back when you're done. 
Could also implement a keybind that makes all of the notes go transparent and be non-interactable so that you can get to the arc/chain nodes, but that feels kind of excessive.

SuperSmay:
I'd imagined the end point nodes as smaller than the notes, so the notes in the same spot would always take priority

CCDV2:
> > But normally there will be a note at end point. It cannot tell you want to drag the note or drag the slider.
> 
> That is true, but if you need to move the end point it seems likely that you either moved the note already or don't have a note there at all. If there is one though you could simply move the note out of the way then put it back when you're done. Could also implement a keybind that makes all of the notes go transparent and be non-interactable so that you can get to the arc/chain nodes, but that feels kind of excessive.

Actually what I'm doing now is adding two note, adding slider and deleleting two notes. A little complex but I think alternative solution seems not to be much better...

SuperSmay:
> Actually what I'm doing now is adding two note, adding slider and deleleting two notes. A little complex but I think alternative solution seems not to be much better...

If you want to modify an arc that already exists or just be able to copy/paste them and move them around you currently need to use node editor. It feels like directly manipulating a preexisting arc would be much easier with draggable nodes, and it would let you skip the placing and deleting process for making new ones that aren’t attached to notes you already have.

Assets/__Scripts/UI/SongInfoEditUI.cs
Caeden117: To clarify, my problem is not that you are modifying `Load_MapV3`, although I'm unsure if that has to be a setting to begin with. The problem I have is with modifying `GridTransparency`, since that is a setting which is directly reflected in the mapper. People will be (and already are) angry when their `GridTransparency` value is overwritten.

Caeden117:
> > > Spline height seems to be a constant in-game, with no relation to start-end distance. Therefore I fix height to 3. It now seems similar.
> > 
> > 
> > From my understanding, arcs also have a clamped floor and left/right sides, depending on the environment (ex. Linkin Park is a much more enclosed environment than something like Billie). At some point, we should make these bounds platform-specific.
> 
> I don't think that's what that is about, but I could be wrong. I think environment clamping isn't implemented at all currently.

I have tested this in-game with a map I've made, and I can confirm that every environment (with the exception of the newest environment, and Skrillex for some reason) clamps the arc trails to a box a tiny bit larger than the 4x3 grid.

Assets/__Scripts/UI/SongInfoEditUI.cs
CCDV2: Ah yes. I've removed the overwritting transparency line.

GalaxyMaster2:
> > Make the end points of arcs (and maybe chains) draggable nodes that can also be rotated, much like normal notes. There would be no need to have the arc indicators be draggable themselves as you can adjust them with the scroll keybinds.
> 
> But normally there will be a note at end point. It cannot tell you want to drag the note or drag the slider.

For this you could simply drag both the end point and the note together. If a note is attached to an arc it stands to reason that you'd want the arc to move with the note.

GalaxyMaster2:
> > > > Spline height seems to be a constant in-game, with no relation to start-end distance. Therefore I fix height to 3. It now seems similar.
> > > 
> > > 
> > > From my understanding, arcs also have a clamped floor and left/right sides, depending on the environment (ex. Linkin Park is a much more enclosed environment than something like Billie). At some point, we should make these bounds platform-specific.
> > 
> > 
> > I don't think that's what that is about, but I could be wrong. I think environment clamping isn't implemented at all currently.
> 
> I have tested this in-game with a map I've made, and I can confirm that every environment (with the exception of the newest environment, and Skrillex for some reason) clamps the arc trails to a box a tiny bit larger than the 4x3 grid.

Yeah I meant in this fork, not in game.

JD-The-65th:
> > > > Spline height seems to be a constant in-game, with no relation to start-end distance. Therefore I fix height to 3. It now seems similar.
> > > 
> > > 
> > > From my understanding, arcs also have a clamped floor and left/right sides, depending on the environment (ex. Linkin Park is a much more enclosed environment than something like Billie). At some point, we should make these bounds platform-specific.
> > 
> > 
> > I don't think that's what that is about, but I could be wrong. I think environment clamping isn't implemented at all currently.
> 
> I have tested this in-game with a map I've made, and I can confirm that every environment (with the exception of the newest environment, and Skrillex for some reason) clamps the arc trails to a box a tiny bit larger than the 4x3 grid.

I assume it's because of the Highway. The new environment and the Skrillex Env have no highway, so the clamping doesn't occur. Something to note for future environments that contain a highway.

GalaxyMaster2:
It seems like ctrl+shift+scroll is broken for some reason, unless that's an issue on my end.

GojiCrafter:
can confirm, ctrl+shift+scroll does not work

Caeden117:
what is ctrl + shift + scroll intended to do? i dont think there's anything in ChroMapper bound to that, including new keybinds introduced in this PR.

GalaxyMaster2:
Changes cursor precision in increments of 1, instead of doubling.

Mawntee:
Why are they called "sliders", and not "arcs"?

Seems odd/confusing to just slap in random names on things, especially when one of them is used to replace the slider patterns mappers have already established years ago.

GojiCrafter:
they're called `"sliders"` in the map format

Caeden117:
While I agree we should rename them to match the community-given name, arcs/chains are still internally referred to as "sliders" and "burst sliders"

Touqai:
Cannot do CTRL + SHIFT + Scroll to precisely change precision, I can only do CTRL + Scroll

Caeden117:
Did a little bit of investigating:

Internally, ChroMapper has a dedicated modifier action bound to `Shift` that toggles whether or not scroll precision will be precisely changed or not. This is used in conjuction with regular `Ctrl + Shift` to change scroll precision by +/- 1.

However, because this PR introduces a new action bound to `Shift + Scroll`, attempting `Ctrl + Shift + Scroll` will now block that modifier action, as ChroMapper is now attempting to execute `Shift + Scroll` (Blocking occurs [here](https://github.com/Caeden117/ChroMapper/blob/dev/Assets/__Scripts/Input/InputSystemPatch.cs#L129-L146).)

This is unfortunate because the keybind blocking system was *intended* to stop this from happening. In this case, we have no choice but to separate the precise scroll precision into its own action bound to `Ctrl + Shift + Scroll`.

CCDV2:
> Did a little bit of investigating:
> 
> Internally, ChroMapper has a dedicated modifier action bound to `Shift` that toggles whether or not scroll precision will be precisely changed or not. This is used in conjuction with regular `Ctrl + Scroll` to change scroll precision by +/- 1.
> 
> However, because this PR introduces a new action bound to `Shift + Scroll`, attempting `Ctrl + Shift + Scroll` will now block that modifier action, as ChroMapper is now attempting to execute `Shift + Scroll` (Blocking occurs [here](https://github.com/Caeden117/ChroMapper/blob/dev/Assets/__Scripts/Input/InputSystemPatch.cs#L129-L146).)
> 
> This is unfortunate because the keybind blocking system was _intended_ to stop this from happening. In this case, we have no choice but to separate the precise scroll precision into its own action bound to `Ctrl + Shift + Scroll` and copy code from AudioTimeSyncController.

OK. Currently it seems that there is no any other precise control. I will make a new method for the interface.

GojiCrafter:
Currently the v3 format created by this fork has 
`"basicEventsTypesWithKeywords": []` but `"basicEventsTypesWithKeywords"` shouldn't be an array, it should have `{}` instead

CCDV2:
> Currently the v3 format created by this fork has `"basicEventsTypesWithKeywords": []` but `"basicEventsTypesWithKeywords"` shouldn't be an array, it should have `{}` instead

Sure. Although `[]` could work in game, but it should be dictionary.

Futuremappermydud:
This is my opinion from what ive seen, I think the arcs should have a better distinction, either with some sort of outline, segment lines, or being thicker. It's really hard to see them especially with lights on.

JD-The-65th:
Funnily enough, I suggested making the arcs thicker back in the (somewhat) early stages. I guess they aren't thick enough.

CCDV2:
> This is my opinion from what ive seen, I think the arcs should have a better distinction, either with some sort of outline, segment lines, or being thicker. It's really hard to see them especially with lights on.

Now they are much shinier...

Futuremappermydud:
Actually, they look pretty good now imo.

Futuremappermydud:
Also the Past Notes Grid does not show chains correctly, only the head as a normal note

CCDV2:
> Also the Past Notes Grid does not show chains correctly, only the head as a normal note

It hasn't been implmeneted.

stbeaumont98:
I'm not sure these will count as bugs, but these are some things I've noticed while using the latest code from this fork:

1. The key mappings for arcs and chains are there, but they don't do anything; or maybe being able to place arcs/chains on the grid hasn't been implemented (?).
2. 'ALT' + dragging arcs or chains doesn't work (again, this could be because it hasn't been implemented yet).
3. Every once in a while, the "head" note of a chain will appear full sized. I wish I could tell you more about why this would happen, but I haven't noticed it being caused by any specific actions on my end. I'll keep an eye out though.

And maybe these are ideas that have already started to be implemented or dismissed for one reason or another, but here are a couple of suggestions that I had:

1. I noticed that some blocks show for the start and the end of the arcs, but they're offset from the actual starts/ends of the arcs (unless the 'mu' and 'tmu' attributes are set to zero). It would be nice if those were always exactly on the starts/ends of the arcs so that, when it is implemented, we could drag the ends to where we want them to be.
2. It would be nice if you could see the arc that is selected. I don't know if it's possible, but maybe the arcs could have an outline or something to indicate they're selected? The start and end "blocks" are highlighted, but because of the offset, they're above and below the map, so they're not always visible.
3. For the chains, a possible implementation could be a draggable icon at the head and tail ends for each's x and y values.

These are just the thoughts I've had so far working with this fork, and I understand if my ideas/opinions are different from anyone else's. And I appreciate all the hard work that goes into this!

JD-The-65th:
> I'm not sure these will count as bugs, but these are some things I've noticed while using the latest code from this fork:
> 
> 1. The key mappings for arcs and chains are there, but they don't do anything; or maybe being able to place arcs/chains on the grid hasn't been implemented (?).
> 2. 'ALT' + dragging arcs or chains doesn't work (again, this could be because it hasn't been implemented yet).
> 3. Every once in a while, the "head" note of a chain will appear full sized. I wish I could tell you more about why this would happen, but I haven't noticed it being caused by any specific actions on my end. I'll keep an eye out though.
> 
> And maybe these are ideas that have already started to be implemented or dismissed for one reason or another, but here are a couple of suggestions that I had:
> 
> 1. I noticed that some blocks show for the start and the end of the arcs, but they're offset from the actual starts/ends of the arcs (unless the 'mu' and 'tmu' attributes are set to zero). It would be nice if those were always exactly on the starts/ends of the arcs so that, when it is implemented, we could drag the ends to where we want them to be.
> 2. It would be nice if you could see the arc that is selected. I don't know if it's possible, but maybe the arcs could have an outline or something to indicate they're selected? The start and end "blocks" are highlighted, but because of the offset, they're above and below the map, so they're not always visible.
> 3. For the chains, a possible implementation could be a draggable icon at the head and tail ends for each's x and y values.
> 
> These are just the thoughts I've had so far working with this fork, and I understand if my ideas/opinions are different from anyone else's. And I appreciate all the hard work that goes into this!

You have to select 2 objects of a note type in order to use the keybinds. Also, make sure you converted your map.

CCDV2:
> I'm not sure these will count as bugs, but these are some things I've noticed while using the latest code from this fork:
> 
> 1. The key mappings for arcs and chains are there, but they don't do anything; or maybe being able to place arcs/chains on the grid hasn't been implemented (?).
> 2. 'ALT' + dragging arcs or chains doesn't work (again, this could be because it hasn't been implemented yet).
> 3. Every once in a while, the "head" note of a chain will appear full sized. I wish I could tell you more about why this would happen, but I haven't noticed it being caused by any specific actions on my end. I'll keep an eye out though.
> 
> And maybe these are ideas that have already started to be implemented or dismissed for one reason or another, but here are a couple of suggestions that I had:
> 
> 1. I noticed that some blocks show for the start and the end of the arcs, but they're offset from the actual starts/ends of the arcs (unless the 'mu' and 'tmu' attributes are set to zero). It would be nice if those were always exactly on the starts/ends of the arcs so that, when it is implemented, we could drag the ends to where we want them to be.
> 2. It would be nice if you could see the arc that is selected. I don't know if it's possible, but maybe the arcs could have an outline or something to indicate they're selected? The start and end "blocks" are highlighted, but because of the offset, they're above and below the map, so they're not always visible.
> 3. For the chains, a possible implementation could be a draggable icon at the head and tail ends for each's x and y values.
> 
> These are just the thoughts I've had so far working with this fork, and I understand if my ideas/opinions are different from anyone else's. And I appreciate all the hard work that goes into this!

Currently you could select arc/chain and cut copy them. Currently I'm busy but I will think of implementing dragging weeks later.

Touqai:
Having an issue where the application starts lagging gradually over time.

PnfrlEnm:
Selecting arcs is (to put it lightly, sorry lol) a royal pain in the ass. Maybe make it so the actual arc itself is clicked on instead of some barely visible block in an arbitrary position that sometimes spawns below the grid so you can't see or click it at all...

Also, the node editor still shows v2 code for regular notes and walls.

JD-The-65th:
> Selecting arcs is (to put it lightly, sorry lol) a royal pain in the ass. Maybe make it so the actual arc itself is clicked on instead of some barely visible block in an arbitrary position that sometimes spawns below the grid so you can't see or click it at all...
> 
> Also, the node editor still shows v2 code for regular notes and walls.
> 
> Big ups to you for implementing v3 this quick, we all owe ya one ;)

I second this, arcs themselves need to be touched in order to select, not the Mu and Tmu blocks. With the Mu and Tmu blocks, it can cause conflict with selecting actual notes

Caeden117:
> > Selecting arcs is (to put it lightly, sorry lol) a royal pain in the ass. Maybe make it so the actual arc itself is clicked on instead of some barely visible block in an arbitrary position that sometimes spawns below the grid so you can't see or click it at all...
> > Also, the node editor still shows v2 code for regular notes and walls.
> > Big ups to you for implementing v3 this quick, we all owe ya one ;)
> 
> I second this, arcs themselves need to be touched in order to select, not the Mu and Tmu blocks. With the Mu and Tmu blocks, it can cause conflict with selecting actual notes

How about, instead of floating blocks for Mu/Tmu, they are instead arrows pointing away from the points? Would make it easier to tell what they are supposed to do, and also makes it easier to select arcs and modify Mu/Tmu

PR #384, opened 2022-04-03
Inverted colors now respect boost lights

XAce1337manX:
In BS 1.21.0 (maybe 1.20.0 as well), inverted colors now respect boost colors.


Caeden117 (approved):
k

PR #387, opened 2022-04-08
Fix create zip button adding Info.dat twice

XAce1337manX:
@Fernthedev


Fernthedev (approved):
I forgot to remove that when doing the chonk merge from upstream, apologies (once again)

Caeden117 (approved):
You're lucky internet out here is barely fast enough to handle github

PR #388, opened 2022-04-10
change v3 custom data key from cd to customData.

CCDV2:
Change custom data key in each object from `cd` to `customData`. Chroma notes should work in v3 map.

Since chroma currently doesn't support colorizing arc and chain, those logic remain unimplemented.


Caeden117 (approved):
LGTM.

PR #390, opened 2022-04-11
Basic function of V3 Walls

CCDV2:
Some commits of walls have been merged. So I will add some introduction of what I've done for v3 obstacles.

- In v3 map, walls will display in v3 height(max at 4.2f). 

However, placement height is still using v2 height(3.5f). Therefore you will experience height difference between placing and placed. Currently I'm not certain about how should we make placement and we may not want to change 3.5f at everywhere, so temporarily keep it for a while..

- To change upper bound, use `shift + scroll`. To change lower bound, use `alt + shift + scroll`.

This is not intuitive for a v3 map wall placing. However, for v3 map, we need to know start height(range from [0, 2]) and height(range from [1, 5]). If we use three step placing(lower bound -> upper bound -> duration), then it is not in cosistent with v2 placement style, because v2 wall's upper bound is fixed. If we use two step placing, then we need to tweak upper bound afterwards. I'm not sure which to choose..

### DIFF
--- a/Assets/Input/Master.cs
+++ b/Assets/Input/Master.cs
@@ -2454,8 +2454,8 @@ public class @CMInput : IInputActionCollection, IDisposable
                 },
                 {
                     ""name"": ""Keyboard"",
-                    ""id"": ""4f8d3770-bbac-4f08-a010-71389fd120e1"",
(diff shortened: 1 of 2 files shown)


Caeden117 (approved):
LGTM

PR #392, opened 2022-04-11
Add floatValue support

XAce1337manX:
`Visualize Chroma Alpha` height accounts for both Chroma alpha and Vanilla floatValue. I don't see a big reason to have these as separate options. As such, I have renamed this option to `Visualize Event Alpha`.

Known behaviour
* Quick editing floatValue does not offset event base to account for height change 
  * This is only done on actions which refresh like node editor or paint


Caeden117 (approved):
LGTM

PR #393, opened 2022-04-14
Add reaction time display in song info

XAce1337manX:
Featuring empty space. (*Please suggest for what can be added to that.*)
[Preview](https://cdn.discordapp.com/attachments/892494185909870602/963746220336103444/RT_field.mp4)


Caeden117:
code seems alright, but if we cant find a 4th thing to fit there, it might be better to move it down and share the row with HJD / Jump Distance (especially since the numbers dont take a whole lot of space anyways)

XAce1337manX:
How about I just...
image

PR #396, opened 2022-04-23
Allow disabling ADB if found in path (also fix downloading)

Fernthedev:
Not tested:tm:

_sigh_


Assets/__Scenes/04_Options.unity
```diff
@@ -18932,7 +19157,7 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   BindedSettingSearchType: 3
-  BindedSetting: UIScale
+  BindedSetting: None
```
XAce1337manX, line 19160: Looks like Unity editor did a funny.
```suggestion
  BindedSetting: UIScale
```

PR #397, opened 2022-04-24
Settings fixes (V3)

XAce1337manX:
Because I broke CameraFOV with floatValue PR.


XAce1337manX:
Closing this as first commit can instead apply to dev while the other commit is entirely on v3.

PR #400, opened 2022-04-29
Arc, chain, and angleOffset improvements

XAce1337manX:
* Support dot arcs
* Chains should be fairly accurate
* Improvements to angleOffset usability
  * Default hotkey is `Shift + Alt + Scroll`


XAce1337manX:
Title change because now has arc improvement. 
(Supports dot arcs.)

PR #404, opened 2022-05-07
Decouple formatted json from custom events setting

XAce1337manX:
New setting in general settings:
image

And also a settings fix


XAce1337manX:
#407

PR #425, opened 2022-09-27
Fix nullref when undoing custom diff name

AllPoland:
Reported in Chromapper server at https://discord.com/channels/702230714585710602/702232127755780197/1024384510189654016


XAce1337manX (approved):
k.

PR #436, opened 2022-11-24
Add feature support up to 3.2.0; Add The Weeknd Environment

CCDV2:
Another fat update on V3 branch.
# 3.1.0 & 3.2.0 Feature Support    
    
Support 3.1.0 features, including:
- More filter options: chunk, random and limit
- Random seed and function may not be exactly the same as ingame

Support 3.2.0 features, including:
- Light translation event
- Easing functions on light/rotation/translation distribution. 

# The Weeknd Environment Support
Build a roughly similar The Weeknd platform, with all the events supported(except ambient light).

# Some Other Changes
- Adjust Light V3's UI
- Add indicator indicating v3 lane's property(rotatable, translate axis, etc.)
- Refractor on V3 LightEvent datas, containers and collections. Now they use template to reduce redundant code.

# Features could be improved
- Ambient light(e.g. in the Weeknd and EDM) is not implemented
- Some following lights are not implemented(e.g. side buildings' lights in the Weeknd, side panels in Fall Out Boys) 

P.S. I will make a map using new environment in the future, but if there is already any map using The Weeknd, please let me know to have a test.


CCDV2:
**Some small bug fix after 2 months**
- Now it can middle click/alt+scroll on v3 events.
- V3 events' texts will follow DisplayFloatText. It can be turned of if there are too much texts.
- Right window order in The Weeknd platform are mirrored. It should display correctly now.

PR #438, opened 2022-12-10
fix broken channel averaging for spectrogram generation

GalaxyMaster2:
previously only the right channel would be used
averaging waveform data isn't technically correct but it's cheap and good enough most of the time
scale factor of 1/1.5 seems to roughly match old behavior from brief testing


XAce1337manX (approved):
Nice.

PR #439, opened 2022-12-11
BPM change handling improvements

GalaxyMaster2:
- makes copy/pasting around bpm changes reliable
- scales wall duration with bpm changes on copy/paste
- accounts for bpm changes when shifting selected objects (shift+arrows)
- updates wall duration tweak to use same bpm change handling logic


XAce1337manX (approved):
LGTM.

PR #444, opened 2023-01-19
Fix and update test

KivalEvan:
title.

## To be done
- [x] Fix test
- [x] Test new v3 stuff
- [x] Test both v2 and v3 objects
- [x] More descriptive test message
- [ ] Another functionality test?
- [x] Temporarily set settings

## Other notable changes & fixes
- Update beatmap factory
- Update slider parameter to be consistent with note **(may break plugins)**
- Fix incorrect NE coordinate key
- Fix mirroring for lane rotation event
- Potential fix for bomb note, rotation event and boost event occasionally not being saved
- Prevent null error from uninstantiated object for event type for keyword


XAce1337manX (approved):
🦦

PR #449, opened 2023-01-28
Unset custom properties when CustomData nodes are removed

MoistSac:
Right now, neither the CustomData nodes nor the Custom* event properties are particularly authoritative, and the only way to fully remove them is to unset both. This PR makes the CustomData node authoritative, and will overwrite and unset properties on ParseCustom(). Going the other way would be much harder, because JSONNodes would need to be `Remove`d and non-events don't yet have Custom* properties.

Implementation note: the ternary is necessary because of JSONLazyCreator, if there was a 
```cs
public JSONNode? TryGet(string key)
```
this could look like
```cs
CustomLerpType = CustomData?.TryGet(CustomKeyLerpType)?.Value
```
~~force push was for commit signing, don't worry about it~~

### DIFF
--- a/Assets/__Scripts/Beatmap/Base/BaseEvent.cs
+++ b/Assets/__Scripts/Beatmap/Base/BaseEvent.cs
@@ -286,24 +286,25 @@ namespace Beatmap.Base
         protected override void ParseCustom()
         {
             base.ParseCustom();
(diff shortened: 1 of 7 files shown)


XAce1337manX (approved):
Yeh.

PR #451, opened 2023-01-31
Graceful plugin failure

MoistSac:
Right now one incompatible plugin will cause all plugins to not run, even though they show up in the plugin list. This prevents that and puts a helpful message in the log.


XAce1337manX (approved):
Very cool.

PR #453, opened 2023-02-04
Feat: Quest Export Buttons and shortcut

Fernthedev:
This PR refactors some of the export code to allow exporting to Quest in other menus, specifically the mapper editor. This has been something that many users including myself have wanted for some time. Finally got some time to implement this.

While I've tested this, I do want to get a bigger testing sample (as my previous feature implementations have lacked proper polish in the past...)

Additions:
- Button for exporting while in editor (pause menu)
- Shortcut for exporting

~~TODO: Hide export button in pause menu unless ADB is installed~~ - Done

image

### DIFF
--- a/Assets/__Scripts/MapEditor/AutoSaveController.cs
+++ b/Assets/__Scripts/MapEditor/AutoSaveController.cs
@@ -4,7 +4,9 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 using Beatmap.Enums;
+using QuestDumper;
 using SimpleJSON;
 using UnityEngine;
 using UnityEngine.InputSystem;
@@ -12,7 +14,7 @@ using UnityEngine.UI;
 
 public class AutoSaveController : MonoBehaviour, CMInput.ISavingActions
 {
-    public bool IsSaving => savingThread != null && savingThread.IsAlive;
+    public bool IsSaving => savingThread != null && !savingThread.IsCompleted;
 
     private const int maximumAutosaveCount = 15;
     [SerializeField] private Toggle autoSaveToggle;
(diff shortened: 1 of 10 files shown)


Assets/__Scripts/MapEditor/MapExporter.cs
```diff
@@ -0,0 +1,187 @@
...
+    private readonly BeatSaberSong song;
+
+    public MapExporter(BeatSaberSong song) => this.song = song;
+
+    /// <summary>
+    /// Exports the files to the Quest using adb
+    /// </summary>
+    public async Task ExportToQuest()
+    {
+        var (devices, output) = await Adb.GetDevices();
```
XAce1337manX, line 35: If the user does the `OnQuestSave` keybind, we don't check if Adb is installed before here so an `InvalidOperationException` can happen.

Assets/__Scripts/MapEditor/AutoSaveController.cs
```diff
@@ -90,6 +92,19 @@ public void OnSave(InputAction.CallbackContext context)
...
+    {
+        if (context.performed) CheckAndSaveQuest();
+    }
+
+    public void CheckAndSaveQuest(int _) => CheckAndSaveQuest(); // So it shows up in Unity
+
+    public async void CheckAndSaveQuest(SaveType saveType = SaveType.None)
+    {
+        CheckAndSave();
+        await Exporter.ExportToQuest();
```
XAce1337manX, line 105: Note that saving is done on a separate thread. The concern is could this attempt to export while file is still being written to?

XAce1337manX (changes requested):
Also worth replacing hyperPepega logError in `CheckAndSave` to something else as this happens everytime the user saves as `OnSave` and `OnSaveQuest` overlap as mentioned above.

XAce1337manX (approved):

Issue #446, opened 2023-01-22
"Export to Quest" button doesn't work on (Debian) Linux


TheZipCreator:
**ChroMapper version**
0.8.523

**Describe the Issue**
Upon clicking "Export to Quest", an error is thrown and the progress is stuck at 0%

**Provide Screenshots or Videos**
https://paste.ee/p/zWMA9 generated bug report

**Steps to Reproduce**
* Install ADB via chromapper
* go to `/path/to/chromapper/quest-utils/platform-tools`
* chmod +x all executables in that directory (this should be done by default, and not doing this causes a different error)
* open a map in chromapper and press "Export to Quest"

**Expected Behavior**
the button should actually work

**Additional Information**
* This button worked on windows before I switched to linux

Fernthedev:
Taking a glance at this, it seems to suggest you’re not using the Linux binary of CM? Why is it throwing Win32 exceptions on Linux? Wine?

That or it’s just the way Mono works on Linux?

also about the chmod +x error, that shouldn’t happen unless extracting the zip loses its permission attributes 🤔

TheZipCreator:
yeah, I was confused about the "Win32Exception" too, since I am using the linux binary. I guess that's just the way Mono works.
also not sure why the permissions aren't saved, does the zip file that chromapper downloads have them? does the library you're using look at file permissions?

MoistSac:
I don't have a quest to test with, but it seems that the adb command to get the device id worked, but the command to create a folder failed. The error log doesn't tell us what error the command made, so can you, from a command line, run:
```
/home/z/things/chromapper/chromapper/quest-utils/platform-tools/adb -s 1WMHH810UM0451 shell mkdir "\"sdcard/ModData/com.beatgames.beatsaber/Mods/SongLoader/CustomWIPLevels/The President is Dead\" -p -m 770
```
and see what the actual error is?

TheZipCreator:
That seems to work fine. Something of note is that in the shell command, the quote is unbalanced. I'm not sure if this is actually a problem or not, but I had to fix it when typing in the command in bash

ChrisJAllan:
Ah yup, [this function](https://github.com/Caeden117/ChroMapper/blob/fbf5a83789333346917cc99b2d0d90459eff845e/Assets/__Scripts/Singletons/Adb.cs#L205) generates unbalanced quotes. That should break export on all platforms, not just linux.  I'll let someone who can actually test if fix it.

[...]

PR #458, opened 2023-02-19
ADB permissions on Mac/Linux

Fernthedev:
It's been made apparent that for whatever reason the executable permissions required to invoke `adb` on Mac or Linux are lost when extracting the zip. 
See #446's steps

The zip does have the permission attributes as shown here: (windows and mac respectively)
image
image

This also cleans up a memory leak when downloading and reading the zip.

I haven't tested this on my Linux machine, though I'd greatly appreciate if someone could vouch for it.

I'll leave it as a ready PR regardless though.

### DIFF
--- a/Assets/__Scripts/Singletons/Adb.cs
+++ b/Assets/__Scripts/Singletons/Adb.cs
@@ -131,6 +131,38 @@ namespace QuestDumper
                 // Extract our zipped file into this directory.
                 archive.ExtractToDirectory(extractPath);
 
+#if UNITY_STANDALONE_OSX || UNITY_STANDALONE_LINUX
+
+                // https://stackoverflow.com/a/63018212/9816000
+                void Exec(string cmd)
+                {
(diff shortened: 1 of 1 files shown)


Assets/__Scripts/Singletons/Adb.cs
```diff
@@ -114,20 +118,51 @@ public static IEnumerator DownloadADB([CanBeNull] Action<UnityWebRequest> onSucc
...
+                    };
+
+                    process.Start();
+                    process.WaitForExit();
+                }
+
+                // Screw it, we're setting the permissions manually
+                // The zip does have the permission attributes, though it seems either Mono or the ZipArchive class
+                // does not properly set them
+                // Obligatory .NET 6 wen
```
Caeden117, line 161: lmao this is unity we're talking about, .net 6 wont be for yearse

XAce1337manX (approved):
Forgot about this. LGTM though I don't have a way to test this.

PR #461, opened 2023-03-01
Optimize gradient displays

Caeden117:
Instead of creating a 256x1 texture for every single gradient, we just do gradient math on the shader. Gradients should now be much faster to initialize, and cooperate with batching as well which should further improve GPU performance.

Only tested with Chroma 2.0 gradients, vanilla transition events remain untested (however I am fairly confident they work well)

EDIT: Also untested is whether or not the GPU easing functions are accurate to the CPU easing functions. I converted the C# easing functions to shader code by hand, and I am well aware that the GPU compiles things differently. Things looked fine when eyeballing comparisons, but that's the only thing I've done.


Assets/_Prefabs/MapEditor/Resources/Event.prefab
```diff
@@ -365,7 +365,8 @@ GameObject:
   serializedVersion: 6
   m_Component:
   - component: {fileID: 4253181293361086480}
-  - component: {fileID: 1018352457705622270}
+  - component: {fileID: 8538710720271473794}
+  - component: {fileID: 1211218263683256254}
   m_Layer: 9
   m_Name: Gradient Sprite
```
XAce1337manX, line 371: It's not a sprite any more. :D

XAce1337manX (approved):
Very cool.

PR #463, opened 2023-03-26
Obstacle CustomSize optional z fix

XAce1337manX:
Dev currently will default CustomSize `z` to 0 instead of it being optional so this fixes that.

### DIFF
--- a/Assets/__Scripts/Beatmap/Base/Customs/INoodleExtensions.cs
+++ b/Assets/__Scripts/Beatmap/Base/Customs/INoodleExtensions.cs
@@ -27,9 +27,23 @@ namespace Beatmap.Base.Customs
         string CustomKeyTailCoordinate { get; }
     }
 
+    public struct Vector2Or3
+    {
+        public Vector2Or3(float? x, float? y, float? z = null)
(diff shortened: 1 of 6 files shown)


Assets/__Scripts/Beatmap/Base/Customs/INoodleExtensions.cs
```diff
@@ -27,9 +27,23 @@ public interface INoodleExtensionsSlider : INoodleExtensionsGrid
...
+        {
+            this.x = x ?? 0f;
+            this.y = y ?? 0f;
+            this.z = z;
+        }
+
+        public float x { get; }
+        public float y { get; }
+        public float? z { get; }
+    }
```
XAce1337manX, line 42: Null z isn't possible with Vector3 so here we are.
Welcome suggestions for where to put this struct as it being in INoodleExtensions seems iffy.

Caeden117: i vote `Beatmap/Shared` or `Beatmap/Helper`

PR #476, opened 2023-05-21
Reunited Mapping

XAce1337manX:
[Related ChroMapTogether PR](https://github.com/Caeden117/ChroMapTogether/pull/1) in order prevent dev and stable from connecting to each other through room codes and destroying the session.
* CM United Mapping is back!
* Added `ApplicationVersion` to `MapperIdentityPacket`
  * Direct connections will now reject differing versions
    * Stable will not reject dev direct connection
    (I don't think I can really do anything about that)


Caeden117 (approved):
LGTM, merge when this has been tested with the updated ChroMapTogether server

PR #484, opened 2023-09-12
Animations

MoistSac:
I think all the bugs are fixed now


Assets/_Graphics/Shaders/Editor/Note.shader
```diff
@@ -260,15 +262,16 @@ Shader "Custom/Note"
...
 				float opaqueAlpha = UNITY_ACCESS_INSTANCED_PROP(Props, _OpaqueAlpha);
-				
-				float alpha = (isTranslucent >= 1 || IN.rotatedPos.w <= 0) ? translucentAlpha : opaqueAlpha;
+
+				float alpha = (animation < 1 && (isTranslucent >= 1 || IN.rotatedPos.w <= 0)) ? translucentAlpha : opaqueAlpha;
 
 				clip(isDithered(IN.ScreenPosition.xy / IN.ScreenPosition.w, alpha));
 
-				return color; // float4(inputData.bakedGI,1);
+				return color * 2; // float4(inputData.bakedGI,1);
```
Caeden117, line 274: Whats the reason behind multiplying color output by 2?

Assets/__Scripts/Beatmap/Animations/ObjectAnimator.cs
```diff
@@ -0,0 +1,561 @@
...
+        private static float WallClamp(float a)
+        {
+            if (-minWall < a && a < minWall)
+            {
+                return minWall;
+            }
+            return a;
+        }
+
+        // I should never be allowed to use a profiler
```
Caeden117, line 516: *goood... goood....*

Assets/__Scripts/Beatmap/Animations/PointDefinition.cs
```diff
@@ -0,0 +1,285 @@
...
+            public InterpolationHandler Lerp { get; }
+
+            public int CompareTo(PointData other)
+            {
+                return Time.CompareTo(other.Time);
+            }
+        }
+    }
+
+    public class PointDataInterpolators
```
Caeden117, line 185: May be smart to abstract this to a different file

Assets/__Scripts/Beatmap/Animations/PointDefinition.cs
```diff
@@ -0,0 +1,285 @@
...
+
+        public delegate T LerpFunc<T>(T prev, T next, float time);
+        static LerpFunc<float> LinearFloat = new LerpFunc<float>(Mathf.LerpUnclamped);
+        static LerpFunc<Color> LinearColor = new LerpFunc<Color>(Color.LerpUnclamped);
+        static LerpFunc<Vector3> LinearVector3 = new LerpFunc<Vector3>(Vector3.LerpUnclamped);
+        static LerpFunc<Quaternion> LinearQuaternion = new LerpFunc<Quaternion>(Quaternion.SlerpUnclamped);
+
+        public static LerpFunc<T> LinearLerpFunc<T>() where T : struct
+        {
+            // I hate C#
```
Caeden117, line 200: I believe newer versions of C# allow you to switch like this

```cs
return T switch
{
    float n => (LerpFunc<T>)(object)(LinearFloat),
    // ...
}
```

But Unity's Mono runtime doesnt support that language feature 🙃

Assets/__Scripts/Beatmap/Animations/PointDefinition.cs
```diff
@@ -0,0 +1,285 @@
...
+            return typeof(T) switch
+            {
+                var n when n == typeof(float) => (LerpFunc<T>)(object)(LinearFloat),
```
Caeden117, line 203: what are the performance costs of boxing/unboxing like this? if its substantial, we may need to rethink this approach.

Assets/__Scripts/Beatmap/Animations/TrackAnimator.cs
```diff
@@ -0,0 +1,155 @@
...
+            case "_color":
+            case "color":
+                AddPointDef<Color>((ObjectAnimator animator, Color v) => animator.Colors.Add(v), PointDataParsers.ParseColor, p, Color.white);
+                break;
+            case "_time":
+            case "time":
+                AddPointDef<float>((ObjectAnimator animator, float f) => animator.SetLifeTime(f), PointDataParsers.ParseFloat, p, -1);
+                break;
+            }
+        }
```
Caeden117, line 134: considering this seems to share some similarities to `ObjectAnimator.cs`, it may be wise to abstract common functionality into a base `GenericAnimator.cs` class.

Assets/__Scripts/Beatmap/Base/Customs/BaseMaterial.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Collections.Generic;
+using Beatmap.Enums;
+using SimpleJSON;
+using UnityEngine;
+
+namespace Beatmap.Base.Customs
+{
+    public abstract class BaseMaterial : BaseItem
+    {
+        // TODO: Add C.U.M.?
```
Caeden117, line 11: better not forget about this TODO (doesnt have to be in this pr)

Assets/__Scripts/Beatmap/Containers/GeometryContainer.cs
```diff
@@ -0,0 +1,89 @@
...
+            container.Shape.transform.localScale = 1.667f * Vector3.one;
+            container.Animator.SetGeometry(eh);
+            container.gameObject.SetActive(true);
+            return container;
+        }
+
+        /// <summary>
+        /// https://answers.unity.com/questions/1594750/is-there-a-premade-triangle-asset.html
+        /// </summary>
+        private static Mesh CreateTriangleMesh()
```
Caeden117, line 59: might be smart to just recreate this triangle in Blender so we never have to touch meshes ourselves

Assets/__Scripts/Beatmap/Containers/NoteContainer.cs
```diff
@@ -11,6 +13,7 @@ public class NoteContainer : ObjectContainer
 
         [SerializeField] private GameObject simpleBlock;
         [SerializeField] private GameObject complexBlock;
+        [SerializeField] public Transform directionTarget;
```
Caeden117, line 16: me when `.editorconfig` violation (you'll have to use `FormerallySerializedAs` attribute to not break references in the Unity prefab/scene)

Assets/__Scripts/Beatmap/Containers/NoteContainer.cs
```diff
@@ -19,13 +22,19 @@ public class NoteContainer : ObjectContainer
         [SerializeField] private SpriteRenderer swingArcRenderer;
 
         [SerializeField] public BaseNote NoteData;
+        public MaterialPropertyBlock arrowMaterialPropertyBlock;
```
Caeden117, line 25: me when the `.editorconfig`

Assets/__Scripts/Beatmap/V2/V2Difficulty.cs
```diff
@@ -301,13 +303,12 @@ private static void LoadCustom(V2Difficulty map, JSONNode mainNode)
                                     foreach (JSONNode n in cNode) bookmarksList.Add(new V2Bookmark(n));
                                     break;
                                 case "_customEvents":
-                                    foreach (JSONNode n in cNode) customEventsList.Add(new V2CustomEvent(n));
+                                    foreach (JSONNode n in cNode) try { customEventsList.Add(new V2CustomEvent(n)); } catch (Exception e) { }
```
Caeden117, line 306: are there exceptions happening here? should probably fix the error at hand rather than silently swallowing it.

otherwise, please format this into multiple lines ty

Assets/__Scripts/MapEditor/UI/UIMode.cs
```diff
@@ -126,17 +132,28 @@ public void SetUIMode(int modeID, bool showUIChange = true)
...
             case UIModeType.Preview:
+                HideStuff(false, false, false, false, false);
+                playerCamera.enabled = false;
+                editorCamera.enabled = true;
+                break;
             case UIModeType.Playing:
                 HideStuff(false, false, false, false, false);
-                OnPlayToggle(atsc.IsPlaying); // kinda jank but it works
+                playerCamera.enabled = true;
+                editorCamera.enabled = false;
```
Caeden117, line 156: could simplify all of this playerCamera/editorCamera assignments to

```cs
playerCamera.enabled = SelectedMode == UIModeType.Playing;
enabledCamera.enabled = !playerCamera.enabled; // Or the inverse SelectedMode comparison as a micro-optimization
```

Caeden117 (commented):
got some room left to cook but i like what im seeing here

Assets/_Graphics/Shaders/Editor/Note.shader
ChrisJAllan: https://discord.com/channels/702230714585710602/892494185909870602/1121697679391596586

With arrows using the note shader (for dissolve), they looked very gray. Multiplying by 2 makes the arrows look white, and makes the notes slightly more vibrant.

Assets/__Scripts/Beatmap/Animations/PointDefinition.cs
MoistSac: Found a different, maybe cleaner way to do this in 26f57e2.

Performance impact of any of these are negligible, old:
Screenshot from 2023-09-17 10:05:34
vs new:
Screenshot from 2023-09-17 10:08:34

The actual lerp function call only accounts for 0.05ms of both.

Assets/__Scripts/Beatmap/Containers/NoteContainer.cs
MoistSac: *My boring text editor doesn't understand .editorconfig and neither do I*
*Also there's no old name so what would `FormerallySerializedAs` point to?*

Caeden117: [`.editorconfig`](https://editorconfig.org/) enforces code styling solution-wide. Makes the code consistent, even when multiple people work on the project. Moreover, because I'm an evil person and want the codebase to at least *look* consistent, [incoming PRs are required to follow `.editorconfig`](https://github.com/Caeden117/ChroMapper/blob/master/CONTRIBUTING.md#coding-convention--styling-guidelines).

Currently, the Unity prefab is looking for `directionTarget` to assign the Transform to. After `.editorconfig` violations have been fixed, this association is broken. You use the `[FormerallySerializedAs]` attribute with the old name of the field:

```cs
[SerializeField, FormerallySerializedAs("directionTarget")] public Transform DirectionTarget;
```

MoistSac: So is the problem that public members must be capitalized? I got a trail of Jetbrains Rider and it still isn't telling me anythings wrong.

Caeden117: yes. interestingly, Rider should be picking up `.editorconfig` automatically.

MoistSac: The rule is there, it just doesn't show any error or warning.
image

Made sure this was turned on too:
image

XAce1337manX:
Oh look, new merge conflicts.
Hopefully this should be the last time you'll need to fix a merge conflict or rebase.

MoistSac:
Rebase done

Also, @Caeden117 one of your commits, b1adfee, deleted the changelog file `q`, was this intended?

XAce1337manX:
Wall appearance is bugged. It appears to be constant size regardless of editor scale.
image
image

MoistSac:
Fixed the wall thing

PR #491, opened 2023-11-01
NAudio support for ChroMapper

Elecast2:
Add NAudio with NVorbis support for better song waveform manipulation and improved delays.

TO DO:
- Add song speed compatibility (Using a custom WaveProvider?)
- Clean code, maybe use a separated file for anything NAudio related
- Get volume from Master, currently it updates volume from song volume but doesn't multiply with master


Assets/__Scripts/MapEditor/AudioTimeSyncController.cs
```diff
@@ -347,27 +376,36 @@ public void TogglePlaying()
...
             {
                 ResetTime();
             }
 
             playStartTime = CurrentSeconds;
-            SongAudioSource.time = CurrentSeconds;
-            SongAudioSource.Play();
+            //ELECAST TEST SongAudioSource.time = CurrentSeconds;
+            //ELECAST TEST SongAudioSource.Play();
+            SetVorbisPosition((long)(CurrentSeconds * bytesPerSecond));//ELECAST TEST
```
GalaxyMaster2, line 388: seeking based on byte position instead of time might be inaccurate with variable bitrates
is there no way to seek based on time?

GalaxyMaster2:
I'm concerned about how this will impact hitsound/metronome synchronization. It might be necessary to handle those through NAudio as well

Assets/__Scripts/MapEditor/AudioTimeSyncController.cs
Elecast2: variable bitrates is not a problem since bytesPerSecond is obtained when we load the song file, or do you mean something else? I couldn't find a way to seek based on time

Elecast2:
> I'm concerned about how this will impact hitsound/metronome synchronization. It might be necessary to handle those through NAudio as well

Eventually the best would be to handle all through NAudio, it will require to write a mixer from scratch tho. Still I think it is more important to find a way to make song speed work before doing all of that.

Assets/__Scripts/MapEditor/AudioTimeSyncController.cs
GalaxyMaster2: bitrate can vary throughout one file, not just between different files
by basing on the average bitrate, you can end up seeking to the wrong time

Elecast2: Oh this makes sence, will check into that

Issue #416, opened 2022-08-01
[Bug] Timeline scrolls on click


TernaryC:
**ChroMapper version**
v0.8.454

**Describe the Issue**
When clicking, the timeline will jump forward or back two intervals; one when pressing the mouse button, and one when releasing (this does not occur when right-clicking, resulting in only one jump). 
The direction seems to be dependent on which direction was most recently scrolled.
Needless to say, this is very irritating.

**Provide Screenshots or Videos**
Video provided showcasing the bug. Note the keylogger in the bottom right, which shows that I am not using the mousewheel when the bug occurs.

**Steps to Reproduce**
Left, middle, or right-click anywhere (not within any UI) shortly/immediately after scrolling

**Additional Context**
OS: Linux Mint 20.3 Cinnamon

https://user-images.githubusercontent.com/36998960/182120917-d35649e4-0a17-46c6-a32f-92bf68b3f554.mp4

JD-The-65th:
Ah. The Linux bug. This is a known issue that as far as I know is unfixable at the moment.
You can try to keep on using it, but if it bothers you Too much, you may just have to use Proton or spin up a Bottle to use the windows version of CM. The bug doesn't happen there.

TernaryC:
Ah, well if it's a known issue I suppose all I can do is deal with it. I'd use the windows version, but it doesn't seem to load through base Wine, Proton, or my software Bottle. If you (or anyone) has any tips for getting the windows version running on linux, that'd be grand, but I know that's pretty nonstandard.

JD-The-65th:
> Ah, well if it's a known issue I suppose all I can do is deal with it. I'd use the windows version, but it doesn't seem to load through base Wine, Proton, or my software Bottle. If you (or anyone) has any tips for getting the windows version running on linux, that'd be grand, but I know that's pretty nonstandard.

If you're launching via CML, that won't work. You have to launch the actually ChroMapper executable. This does mean you'll have to manually update ChroMapper. No worries, however. We have a Jenkins for this so that you can just grab the newest version whenever you need it!
https://jenkins.kirkstall.top-cat.me/view/All/job/ChroMapper/

PR #493, opened 2023-11-05
Update to Unity 2021.3

Caeden117:
This PR updates the Unity version of ChroMapper to **Unity 2021.3.32f1**.

This Unity update brings updates to several packages, as well as a switch to **.NET Standard 2.1** so newer features of .NET can be used in the project. The change to .NET Standard is important for future additions I want to make to the project ([AudioSync](https://github.com/Caeden117/AudioSync), the BPM detection library I wrote, heavily relies on `Span<T>`, which was added in .NET Standard 2.1)

**TODO:**
- Investigate: Input System update might've fucked up key rebinding


XAce1337manX:
Bonus: I believe this may fix #416.

PR #494, opened 2023-11-07
BPM Detection

Caeden117:
(Requires #493)

This PR implements a method of BPM detection to ChroMapper using [AudioSync](https://github.com/Caeden117/AudioSync), a pure C# implementation of ArrowVortex's BPM detection algorithm.

Results should be pretty comparable to ArrowVortex, however the speed of AudioSync seems to be noticeably slower than in my basic benchmarking on .NET 7. I suspect this may be due to (i) Unity's Mono runtime or (ii) using Development/Debug mode for ChroMapper.

**To-do List:**
- [x] Implement AudioSync
- [ ] Show more details about results in completion dropdown (Offset + Fitness?)
- [ ] (more of an AudioSync thing) Add progress reporting to bpm detection
- [ ] Apply detected BPM + offset to the map
- [ ] Localize dialog boxes
- [ ] Reference AudioSync via NuGet (NuGetForUnity was added in the 2021.3 PR) once AudioSync is published to NuGet


GamesGG0:
Are yall still working on this?

PR #500, opened 2024-01-05
Disable Harmony patches on macOS

Caeden117:
This PR aims to restore ChroMapper compatibility on Apple Silicon devices by disabling its Harmony patches on macOS.

On current `dev`, this may lead to problematic behavior relating to the Input System, which relies heavily on its Input System patch. However, on `2021.3`, this Harmony patch may be less necessary. If needed, this PR can be re-targeted to `2021.3` and merged there.

Unfortunately, Better Bloom will be completely disabled on macOS, since Universal Render Pipeline still has no option to enable a full resolution bloom effect.

Also to note, Unity does not seem to define specific preprocessor symbols for M1 specifically, so these changes will apply to Intel based Apple devices as well, even though Harmony works just fine on Intel devices.

### DIFF
--- a/Assets/__Scripts/Input/InputSystemPatch.cs
+++ b/Assets/__Scripts/Input/InputSystemPatch.cs
@@ -21,6 +21,7 @@ using UnityEngine.InputSystem.Controls;
  */
 public class InputSystemPatch : MonoBehaviour
 {
+#if !UNITY_STANDALONE_OSX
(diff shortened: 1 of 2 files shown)


XAce1337manX (approved):
This seems to work. A quick test build for Apple Silicon isn't immediately crashing on startup for M1 mac.

PR #501, opened 2024-01-10
Animation fixes

MoistSac:
Fixes:
- NJS/JD changing after BPM changes
- Arcs and chains being offset in preview mode
- Placing notes outside of vanilla range at edges


Caeden117 (approved):
looks good to me

PR #502, opened 2024-01-10
Camera fixes

XAce1337manX:
* Fix raycast coming from inactive camera
* Delete tool is only active with Application focus


Caeden117 (approved):
lgtm

PR #504, opened 2024-01-16
Fix Typo; recommendend to recommended

JD-The-65th:
Title is self explanatory. Ironically while making this commit and PR, I misspelt recommended several times.


XAce1337manX:
Btw this looks like a result branching off of master instead of dev.

JD-The-65th:
Weird. I told Github to choose from dev when I entered github.dev but I guess it didn't listen.

PR #506, opened 2024-01-22
Stable (0.9.690)

XAce1337manX:
Nice.

Stable candidate featuring animations, colour options, M1 crash fix, and other goodies.

~~Note 1: Before merging this into master, we may want to remove the Save Without Default Values setting commit on master to avoid duplicate commit.~~ It's a tagged release, I'll just leave it as is.
  
~~Note 2: After merging this into master, we may want to either rebranch dev off of master and fix old code only present in master for some reason (and prevents compiling from fresh clone).~~ May affect 2021.3 branch, so I'll also leave as it. People can rebase changes onto dev.


Caeden117 (approved):
feel free to merge whenever. can't wait to wrap up 2021.3

PR #507, opened 2024-01-26
Better time keeping

Caeden117:
This Pull Request implements a more robust form of updating the time spent in a map. This new version should be more accurate, and less prone to floating point error.


XAce1337manX (approved):
LGTM.

PR #508, opened 2024-02-01
Fix dragging objects not working

XAce1337manX:
* Instead of deleting and spawning on FinishDrag, we delete on StartDrag and spawn on FinishDrag.
* Remove some copying while dragging


Caeden117 (approved):
LGTM

PR #509, opened 2024-02-02
Add Interface Opacity

Caeden117:
while chromapper is currently blowing up with save issues from my own rewrites... how about a simple appearance setting? :)

one of a few requests from mawntee


XAce1337manX:
Two decimals seems like way too much precision for the slider. Any objections if I change it to 0 decimals to match grid transparency?
image

Also doesn't appear in search. Will fix.
image

Caeden117:
i was running into default coloring problems on 0 decimals, but feel free. perhaps i was doing things wrong

XAce1337manX:
Ran into same default colour problem. Turned out to be bug.

XAce1337manX (approved):

PR #510, opened 2024-02-08
Optimizations with Span<T>

Caeden117:
This PR optimizes accelerates various components of ChroMapper by taking advantage of [`Span<T>`](https://learn.microsoft.com/en-us/archive/msdn-magazine/2018/january/csharp-all-about-span-exploring-a-new-net-mainstay). Iterating over a `Span<T>` is noticeably faster, even when compared to iterating a `List<T>` using a `for` loop. In addition, `Span<T>` iteration does not allocate memory, unlike a `foreach` loop.

Made possible by [UnityH4xx](https://github.com/atcarter714/UnityH4xx), a Unity library that uses custom written IL to expose the backing array of a `List<T>` as a `Span<T>`. This is official API in .NET 5 and later, but is unfortunately left out of .NET Standard 2.1.

[The Only Thing They Fear is You](https://beatsaver.com/maps/943e) is a good benchmark map if you want to compare performance improvements in this PR.

ChroMapper code that was touched by this PR:
- `BeatmapObjectCallbackController<T>.RefreshPool` is almost entirely rewritten, not only to utilize `Span<T>` but also using Binary Search to reduce the average runtime from O(n) to O(log n).
  - Pausing, unpausing, and scrolling through any object-heavy level should see noticeable improvements.
- `EnumerableExtensions` methods now take advantage of `Span<T>`.
- `BeatmapObjectCallbackController<T>.GetViewBetween` is now `Span<T>` accelerated, providing a direct view into the `MapObjects` list rather than allocating an entirely new collection.
- `CustomEventsGridContainer` now uses `Span<T>` iteration when it previously used for/foreach iteration
- `RotationCallbackController` sees similar accelerations by using both `Span<T>` and Binary Search
  - Pausing, unpausing, and scrolling through 360/90 levels dense with rotations should see additional improvements.
- `MapLoader` uses `Span<T>` to quickly recalculate song bpm time
- Strobe Generator has been touched a little bit, being *slightly* more memory efficient by using the `Span<T>`-backed `GetViewBetween`
  - However, Strobe Generator itself is still largely inefficient thanks to heavy LINQ abuse. It deserves a proper rewrite in due time.
- While a little out of PR scope, `GridRotationController` no longer uses a coroutine to handle 360/90 rotation transitions. This saves some accumulated garbage every frame.


XAce1337manX (approved):
LGTM.

PR #517, opened 2024-03-23
Fix track filter

MoistSac:
Makes the Shift+T track filter actually work! It hasn't done anything for as long as I've seen.


XAce1337manX (approved):
Didn't even know we had this feature lol. LGTM.

PR #518, opened 2024-03-24
Display SongCore folders in SongSelect

XAce1337manX:
* If a SongCore `folder.xml` exists, they will now be displayed in song select
image

* More folders
image

* No support for adding folders yet


Caeden117 (approved):
Looks good to me as preliminary support. Folder management can come later.

PR #522, opened 2024-04-30
GPU Accelerated Spectrogram

Caeden117:
This PR completely redesigns ChroMapper's spectrogram, now accelerated by the graphics card. The new spectrogram can generate in fractions of time compared to the old one, and can now be completely hidden by the loading screen.

image

New spectrogram settings have been added, including:
- Sample Size
- Logarithmic scale
- Shell texturing count + height

3D Spectrogram has served no functional purpose for years and is officially removed as of this PR.

This also includes a CrowdIn pull since new localization keys had to be made and I was worried about losing potential localizations that haven't yet been pulled.

**Known Issues:**
- ~~Certain songs produce artifacting on the top and bottom edges of the song preview:~~
image
  - ~~These artifacts seem to be resolution dependent, and don't show in the 3D editor~~
- ~~The spectrogram preview also doesn't respect the UI mask, which would hide the aforementioned artifacting~~
  - ~~Fixed mask but artifacting still shows~~
- Cross-platform compatibility has NOT been tested.
  -  Theoretically this should work on any platform that supports Compute Shaders (See [Unity docs](https://docs.unity3d.com/Manual/class-ComputeShader.html), it seems like every platform we are currently targeting has Compute Shader support)


XAce1337manX:
Spectrogram not showing up in editor is probably from the last push I did oops. (Tunneled too hard on the masking.)

However for the other issue, that mystery scrollbar might be the console opening but failing to render everything else for whatever reason. There's hopefully a useful error in that log.

XAce1337manX:
Spectrogram artifacts do actually appear in the 3D editor when using low sample quality settings:
image

MoistSac:
Okay, previous issues seem fixed, not sure if it was the last commit or me forcing Unity to use Vulkan. But I bring a new bug!

The spectrogram seems to go "choppy" over time. It's immediately noticeable on the song info screen, but the editor one goes eventually as well.
Dev:
Screenshot from 2024-05-02 10:54:32
GPU Spectrogram (first time):
Screenshot from 2024-05-02 10:54:34
After viewing other maps:
Screenshot from 2024-05-02 11:01:23

Editor, near end of song:
Screenshot from 2024-05-02 10:55:50
Subsequent song loads:
Screenshot from 2024-05-02 11:01:34

Also yes, lots of graphics errors in the logs. These errors don't show up in the editor, only the build, but the above bug is present in both.
[ChroMapper.log](https://github.com/Caeden117/ChroMapper/files/15192319/ChroMapper.log)

PR #526, opened 2024-05-28
Better Plugin Debuging

MoistSac:
Does two things:

1. Keep loading plugins if one has an error.
2. Actually give useful exceptions for plugins instead of just a `TargetInvocationException`.


Caeden117 (approved):
LGTM

PR #529, opened 2024-06-21
Remove code from master which doesn't exist in dev

XAce1337manX:
Anyone that tries to build master currently is going to run into compile errors due to some code that shouldn't exist: https://github.com/Caeden117/ChroMapper/compare/master..54da68246149794f89517924bc7f343e23903ecb

This makes master match dev. Maybe worth rebranching dev off of master after this merge so the branches are synced again.


Caeden117 (approved):
LGTM

PR #534, opened 2024-07-08
Fix the open config folder button on Linux

MoistSac:
Uses the same behavior as MapExporter and OptionsController.
*This should probably be made into a helper function somewhere if you have a good place to put it.*


Caeden117:
Currently theres no convenient `Utils` folder or something. Probably something like the root `__Scripts` is good enough, then eventually we can move some helper scripts into a more proper Utils folder.

Either way, I do agree that by this point, it should probably be made into a helper function

MoistSac:
Helper function added

Caeden117 (approved):
LGTM

PR #537, opened 2024-07-15
Exclude geometry lights from propid

MoistSac:
Fixes an issue where all geometry-added light ids are grouped under propid 1

Example:
Screenshot from 2024-07-15 13:15:44


Caeden117 (approved):
LGTM

PR #541, opened 2024-07-27
Stable (CM 0.10.779)

Caeden117:
bug fixes that we kinda need plz thx


XAce1337manX:
^ Commit rewrote history to merge with history instead squashing

PR #544, opened 2024-08-24
Refactor base difficulty and map objects

XAce1337manX:
* Changed base classes of map objects to be concrete classes. It's the internal model CM uses for data. (e.g. BaseNote)
* Version specific classes are used only to parse from and to json. 
  * Two methods: `GetFromJson` and `ToJson`. 
  * Any supported currently custom json data keys
* `Settings.Instance.MapVersion` is a global setting for what version to output to
  * This is set only when loading a map and switching version
  * This is what node editor checks to get and parse json

Things left todo:
* Test this more


Caeden117 (approved):
Initial review seems OK. will review further changes once you return o7

PR #552, opened 2024-12-10
v4 Format Support

XAce1337manX:
Features:
* Can edit v4 maps - Mainly for compatibility with official editor
* Saving with incompatible data in format will prompt a warning
* Additions to Song Info screen to support new fields
  * Vanilla: 
    * Mappers, lighters, and shared lightshow fields
    * Environment selection per diff
  * Custom:
    * Read support for custom characteristics
    * Stub wip menu for new SongCore features (lets users know the feature exists)
    * Small UX improvements to pop up menus (rewording and closing)

| SongCore custom characteristic display | Environment Removal moved into Custom Properties menu |
| :-: | :-: |
| image | image |

(If any one wants to offer suggestions to the menu, feel free)

Brief overview of changes:
* BeatSaberSong replaced with BaseInfo. Reading to and writing from this is handled in V2Info and V4Info
  * A majority of changes will be just changing this reference
  * V4 classes, as with V2 and V3, just handle IO
* New maps are created with v4 info as it gives access to some new metadata fields and still supports difficulties of any version.
* New difficulties are still created in v3 as CustomJSONData does not have a specification for v4. For now, v4 custom properties will display with v3 json in editor and will not persist on save

_Lots of things to deal with. I expect to find more bugs when this gets merged._

PR #554, opened 2024-12-14
Stable (CM 0.11.793)

XAce1337manX:
Mostly bug fixes. Want a stable release before merging #552 into dev.
Also getting onto fixing all the plugins that have broken.


Caeden117 (approved):
LGTM, merge whenever

PR #556, opened 2024-12-19
Fix metronome double tick

UGEcko:
Fixed metronome double tick
Also changed some spots to use `metronomeActive` instead of `metronomeVolume`

### DIFF
--- a/Assets/__Scripts/MapEditor/Audio/MetronomeHandler.cs
+++ b/Assets/__Scripts/MapEditor/Audio/MetronomeHandler.cs
@@ -17,30 +17,50 @@ public class MetronomeHandler : MonoBehaviour
     private Animator metronomeUIAnimator;
     private bool metronomeUIDirection = true;
 
-    private float metronomeVolume;
-
+    private bool metronomeActive = true;
+    
     private float songSpeed = 1;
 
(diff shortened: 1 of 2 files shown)


XAce1337manX:
For next time should drop or squash the commits that you don't need. (The second just reverts the first.)

Can squash merge this on the weekend after I merge v4.

UGEcko:
Yeah I apologize, I didnt intend on committing a file that was changed on the first commit, I thought the last commit would be the latest.
Also thank :)

Assets/__Scripts/MapEditor/Audio/MetronomeHandler.cs
```diff
@@ -73,7 +92,7 @@ private void RunAnimation(float inferredBpm)
 
     private void OnPlayToggle(bool playing)
     {
-        if (metronomeVolume == 0) return;
+        if (Settings.Instance.MetronomeVolume == 0) return;
```
XAce1337manX, line 95: Can probably replace this while we're here
```suggestion
        if (!metronomeActive) return;
```

UGEcko:
Oh yea good catch

XAce1337manX (approved):

PR #560, opened 2025-02-20
[macOS] Change App Category to Prevent Game Mode

ztc0611:
The current (Unity default) app category causes macOS to enter game mode when the app is put in full screen, which is not ideal for energy purposes. I am doing a broader investigation of the program preventing sleep and if it's possible to avoid, but this change is still relevant for energy use reduction.

Game mode is controlled by having one of the `game`-related tags set in Mac App Store Options/Category field in Unity. (Maps to `LSApplicationCategoryType`) I've changed it to `developer-tools` but `utilities` could also fit.

Note: Game Mode will still activate when you run the built project on a Mac if Unity Hub is running. This doesn't appear to be the case when it is closed, or sent to a computer without Unity Hub installed.

https://developer.apple.com/forums/thread/739387
https://developer.apple.com/documentation/bundleresources/information-property-list/lsapplicationcategorytype
https://docs.unity3d.com/2020.1/Documentation/Manual/HOWTO-PortToAppleMacStore.html

### DIFF
--- a/ProjectSettings/ProjectSettings.asset
+++ b/ProjectSettings/ProjectSettings.asset
@@ -91,7 +91,7 @@ PlayerSettings:
   useFlipModelSwapchain: 1
   resizableWindow: 1
   useMacAppStoreValidation: 0
-  macAppStoreCategory: public.app-category.games
+  macAppStoreCategory: public.app-category.developer-tools
   gpuSkinning: 0
   xboxPIXTextureCapture: 0
   xboxEnableAvatar: 0


Caeden117:
I do not have an Apple Silicon Mac device to test this change myself. Have you tested this on your own machine? I am OK to merge this change if such is the case.

ztc0611:
I tested it on two Apple silicon machines, worked on them. I'm also not sure if you've changed this, but for whatever reason the current public download is intel only. When I compiled it on my machine it made a universal binary. Is that a limitation of intel machines using Unity?

Caeden117:
ChroMapper should be being built as a universal x86 / ARM binary since #460. We are using CI to compile and distribute the public builds, so if ARM support requires compilation on an ARM machine, we may have to look into that.

ztc0611:
I'm not super familiar with how universal binary compilation works, I never thought it required special hardware. When I go to the linked download page (https://cm.topc.at/dl) the binary that I get says Intel only. (Compared with a VSCode Universal)

Apple's developer [documentation](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary) says you can compile for arm64 on intel machines, so I'm not sure what's wrong. I know Godot compiles a universal binary with not much trouble, but I'm not sure how much of this is down to Unity or some other force I'm unfammilar with. I will look into this further in the next few days, I have an important job interview tomorrow morning.

image
image

Caeden117:
Ah, that seems to be the case for the ChroMapper Launcher, which is a separate executable from the main ChroMapper program. I am unsure how easily CML can be built as a universal library.

Top-Cat:
I don't think it's that important for it to be as long as CM itself is universal (which afaik it is). You're really worried about performance of downloading updates?

ztc0611:
I misunderstood the setup. When I was experimenting with the change I made, I only built the main program and didn't do anything with the launcher. I see that the actual ChroMapper program is deposited in `/User/Applications/` (vs the typical, all-user `/Applications/`) after the launcher does its thing, and the main app is Universal. There is the "some other force I'm unfamiliar with," haha.

Apple might drop support for Rosetta 2 (Intel on Apple Silicon) in the future, but it's probably not worth being concerned with in the here in now.

Now that's clarified (sorry for the confusion), the original game mode part still stands.

Caeden117 (approved):
Feel free to continue your investigation into power efficiency on Apple Silicon. I'll merge this change as-is.

PR #564, opened 2025-03-15
remove known and parsed customData from base CustomData object

GalaxyMaster2:
fixes converted custom bpm changes not being deleted, among other potential similar issues

(now also on Info.dat, fixing issues with removing contributors as one example)

(now also fixed diff label not being copied to new diff, among some other properties)


Caeden117 (approved):
Code looks good, will wait for Bullet's personal testing before merging

XAce1337manX:
And probably related to above issue. Chroma in this case doesn't get removed on save once I remove the Chroma things from the map.

XAce1337manX (approved):

PR #567, opened 2025-03-17
sort maps in songlist based on modified date of Info.dat

GalaxyMaster2:
the modified date of the containing directory only changes when new files are written
instead we can use the modified date of Info.dat directly, which is updated whenever the map is saved


XAce1337manX (approved):
I think it's worth commenting the reason we'll use the Info.dat instead of the directory.

PR #571, opened 2025-03-23
Fix camera controller snapping on first mouse move

Umbranoxio:
Unsure if this is just a macOS issue and no idea if this fix works for everyone, but hey it's something.

The fix I implemented is to just ignore the first few frames of mouse input after right clicking, then gradually ease into normal camera control

Before:

https://github.com/user-attachments/assets/f4d32543-6f9d-479a-be02-4ba6e2c3a165

After:

https://github.com/user-attachments/assets/be5b9a37-7c8d-4eab-a3bb-ea279f44985f


Top-Cat:
I do think that is OSX specific

Recording from v0.12.832 on Windows
https://github.com/user-attachments/assets/74eff5e9-75d4-4f7e-badf-bfde04a88df1

Caeden117 (approved):

PR #572, opened 2025-03-23
various selection fixes

GalaxyMaster2:
* fixed selection outline randomly becoming green
* fixed node editor not updating after object has been dragged
* removed unused node editor enable setting (fixes node editor not updating when nothing is selected, if this setting was disabled)
* fixed desyncs from various undo/redo actions not invoking SelectionChangedEvent
* fixed box selection invoking SelectionChangedEvent too early, which can result in desync


Caeden117 (approved):
LGTM but would also like Bullet's eyes on the PR before merge

XAce1337manX (approved):

PR #579, opened 2025-03-25
BPM event refactor

GalaxyMaster2:
**BPM event refactor:**
* songBpmTime is now nullable. If time conversion fails, it is set to null. Casting it to float when it's null will raise an exception. This is intentional; it means there is a serious bug and we should know about it.
* SongBpmTime cannot be set anymore. This was only used in one place, and you should set JsonTime instead.
* BPM time conversion logic is moved to from BPMChangeGridContainer to BaseDifficulty. BPM events must be bootstrapped after loading for this to function. All object songBpmTimes should be recomputed after doing this.
* SetTimes in BaseObject has been simplified since songBpmTime was getting recalculated anyway.

This fixes wrong saving of BPM info files (BPMInfo.dat/AudioData.dat) when saving diffs from the song edit screen. It should also make it impossible for such a bug to go unnoticed in the future.
I haven't found any new bugs stemming from this refactor, but more testing is of course welcome.

---
**mitigate accumulating numerical errors on BPM region load/save:**
* When saving BPM regions, properly round calculated sample indexes, and ensure continuity by setting start sample index to previous end sample index.
* When converting BPM regions into events, round the BPM value of resulting events if the unrounding was (likely) caused by numerical error.

Together these should (mostly) eliminate accumulating errors on load/save cycles, which is important for v4 maps, since they do not store the actual BPM events.

### DIFF
--- a/Assets/__Scripts/MapEditor/Grid/Collections/BeatmapObjectContainerCollection.cs
+++ b/Assets/__Scripts/MapEditor/Grid/Collections/BeatmapObjectContainerCollection.cs
@@ -358,9 +358,14 @@ public abstract class BeatmapObjectContainerCollection : MonoBehaviour
 
     public static void RefreshFutureObjectsPosition(float jsonTime)
     {
-        foreach (var objectType in System.Enum.GetValues(typeof(Beatmap.Enums.ObjectType)))
+        // we have to refresh bpm events FIRST, and only then can we refresh other objects
+        // this is a janky way of accomplishing that
+        var objectTypes = new List<ObjectType> { ObjectType.BpmChange };
+        objectTypes.AddRange(((IEnumerable<ObjectType>)Enum.GetValues(typeof(ObjectType))).Where(x => x != ObjectType.BpmChange));
+
+        foreach (var objectType in objectTypes)
         {
-            var collection = BeatmapObjectContainerCollection.GetCollectionForType((Beatmap.Enums.ObjectType)objectType);
+            var collection = BeatmapObjectContainerCollection.GetCollectionForType(objectType);
             if (collection == null) continue;
             // REVIEW: not sure if allocation is avoidable
             foreach (var obj in collection.LoadedObjects)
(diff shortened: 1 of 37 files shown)


Assets/__Scripts/MapEditor/Grid/Collections/BeatmapObjectContainerCollection.cs
```diff
@@ -358,9 +358,14 @@ public abstract void SpawnObject(BaseObject obj, out List<BaseObject> conflictin
 
     public static void RefreshFutureObjectsPosition(float jsonTime)
     {
-        foreach (var objectType in System.Enum.GetValues(typeof(Beatmap.Enums.ObjectType)))
+        // we have to refresh bpm events FIRST, and only then can we refresh other objects
+        // this is a janky way of accomplishing that
+        var objectTypes = new List<ObjectType> { ObjectType.BpmChange };
+        objectTypes.AddRange(((IEnumerable<ObjectType>)Enum.GetValues(typeof(ObjectType))).Where(x => x != ObjectType.BpmChange));
```
Caeden117, line 364: At this point, considering we are likely not to add any future `ObjectType`s for quite a while, if at all... would it make more sense just to hardcode this list?

Caeden117 (approved):
LGTM, although PR is big enough for me to want a second look from Bullet before merging

Assets/__Scripts/MapEditor/Grid/Collections/BeatmapObjectContainerCollection.cs
GalaxyMaster2: yeah I guess I'll just do that

XAce1337manX:
Tests are failing.
Also cannot load into the editor on some maps. (e.g. key 44178)

XAce1337manX (changes requested):
Tests are failing.
Also cannot load into the editor on some maps. (e.g. key 44178)

GalaxyMaster2:
tests should be passing now

XAce1337manX (approved):

PR #581, opened 2025-03-27
merge specific actions when repeated

GalaxyMaster2:
quality of life improvement with undo/redo of repeated actions, such as tweaking values with alt+scroll

currently not tested with united mapping, PR is marked as draft until this is done


GalaxyMaster2:
From testing it now seems to work in united mapping. Yes it does break when multiple people have modified the same object and try to undo, but this is a separate issue that already existed before. It's not any more broken than it already was, so I'll consider it something to fix in a future PR.

XAce1337manX (approved):

PR #582, opened 2025-03-31
fix spectro alignment on track length not multiple of 4

GalaxyMaster2:
integer division my behated


Caeden117 (approved):
me when INTEGERS‼️

PR #583, opened 2025-04-07
GAGA

UGEcko:
_Gaga? Is that you?_

Add Lady Gaga Environment

Thank you GamesGG and Caeden for your help! <3


GamesGG0:
Yoooo!!! Gaga!!! The best V2 environment in chromapper!

Assets/__Scripts/Platforms/Gaga/GagaDiskManager.cs
```diff
@@ -0,0 +1,164 @@
...
+    private Dictionary<int, List<BaseEvent>> cachedHeightEvents = new Dictionary<int, List<BaseEvent>>();
+
+    public void Start()
+    {
+        atsc = FindObjectOfType<AudioTimeSyncController>();
+        eventGridContainer = FindObjectOfType<EventGridContainer>();
+        foreach (var disk in Disks)
+        {
+            // Start at Y 20 (default).
+            disk.Init(20);
```
XAce1337manX, line 26: Think this makes sense as default paramter in the init?
`public void Init(float positionY = 20f)`

Assets/__Scripts/Platforms/Gaga/GagaDiskManager.cs
```diff
@@ -0,0 +1,164 @@
...
+    {
+        foreach (var disk in Disks)
+        {
+            disk.LateUpdateDisk(atsc.CurrentJsonTime);
+        }
+    }
+
+    public void HandlePositionEvent(BaseEvent evt)
+    {
+        var value = evt.Value;
```
XAce1337manX, line 52: Unused variable.

Assets/__Scripts/Platforms/Gaga/GagaDiskManager.cs
```diff
@@ -0,0 +1,164 @@
...
+                {
+                    var fromValue = evt.Value;
+                    var toValue = nextEvent.Value;
+                    var toTime = nextEvent.JsonTime;
+                    d.SetPosition(ClampValue(fromValue), ClampValue(toValue), evt.JsonTime, toTime);
+                }
+            });
+    }
+    
+    private int ClampValue(int value) => Math.Clamp(value, minEventValue, maxEventValue);
```
XAce1337manX, line 68: May want to make this name more explicit: `ClampEventValue`

Assets/__Scripts/Platforms/Gaga/GagaDiskManager.cs
```diff
@@ -0,0 +1,164 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Beatmap.Base;
+using UnityEngine;
+
+public class GagaDiskManager : MonoBehaviour
+{
+    private readonly int minEventValue = 0;
+    private readonly int maxEventValue = 8;
```
XAce1337manX, line 10: These can be `const`.

Assets/__Scripts/Platforms/Gaga/GagaDiskManager.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            disk.LateUpdateDisk(atsc.CurrentJsonTime);
+        }
+    }
+
+    public void HandlePositionEvent(BaseEvent evt)
+    {
+        var value = evt.Value;
+        Disks.Where(d => d.HeightEventType == evt.Type)
+            .ToList()
```
XAce1337manX, line 54: Is to `ToList()` necessary as opposed to a `foreach`?

Assets/__Scripts/Platforms/Gaga/GagaArc.cs
```diff
@@ -0,0 +1,40 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class GagaArc : MonoBehaviour
+{
+    public GameObject TargetObject;
+    public Material ArcMaterial;
+    private readonly float thickness = 1f;
```
XAce1337manX, line 10: Can be `const`.

Assets/__Scripts/Platforms/Gaga/GagaArc.cs
```diff
@@ -0,0 +1,40 @@
...
+        lightningMesh.transform.SetParent(gameObject.transform, false);
+        lightningMesh.transform.position = Vector3.zero;
+    }
+    void Update()
+    {
+        // Update plane transform between two points.
+        var sPosition = transform.position;
+        var tPosition = TargetObject.transform.position;
+       
+        float filterLogo = -Math.Sign(tPosition.x - (Math.Sign(tPosition.x) * 14.08));
```
XAce1337manX, line 27: Where does `14.08` come from?

UGEcko: This was used to determine the angle the plane of the lightning is facing. 14.08 is the closest appropriate position thats greater than the 2 inside towers positions.

Assets/__Scripts/Platforms/Gaga/GagaArc.cs
UGEcko: gotcha

UGEcko:
Id like to note here that the gaga shaders dont work display alpha colors on events, its only 1 or 0.
Will look at it again tomorrow

Assets/__Scripts/Platforms/Gaga/GagaArc.cs
XAce1337manX: Should indicate this by a variable name or comment.
Magic numbers are stinky!

Assets/__Scripts/Platforms/Gaga/GagaDiskManager.cs
UGEcko: Good idea

Assets/__Scripts/Platforms/Gaga/GagaDiskManager.cs
UGEcko: Not at all in this case lol

Assets/__Scripts/Platforms/Gaga/GagaDiskManager.cs
XAce1337manX: Yes. There is no point adding code that does nothing. (Issue is already resolved.)

XAce1337manX (changes requested):
Regression: maps that use an environment that CM doesn't have support for (any group lighting system environment) will not load and throws IndexOutOfRangeException. Previous working behaviour was to load DefaultEnvironment.

UGEcko:
Fixed. Sowwy

UGEcko:
Fixed aurora/lightning shaders + removed unnecessary lightIDs.

XAce1337manX (approved):

PR #584, opened 2025-04-11
fix song bpm duration recompute when obstacle time changes

GalaxyMaster2:
and a bit of refactoring while I was at it


GalaxyMaster2:
apparently tests are failing even though they were passing last time (somehow)
will fix later

GalaxyMaster2:
should be good now

XAce1337manX (approved):

Caeden117 (approved):

PR #599, opened 2025-08-30
Fix weird chain head behaviour

KivalEvan:
Chain head note should be much more consistent than having erratic transform placement, performance wise dont see a difference so should be good?

Haven't fully tested it with various scenario, basic interaction seems to be ok
This is for hot fix patch before push for major dev update, these may get rewritten for better implementation


Caeden117 (approved):
Code looks good but I will not be available to test until Wednesday. Bullet plz 🥹🥹🥹🥹

PR #601, opened 2025-09-03
Better chain model

KivalEvan:
Fairly small change for visual upgrade, definitely need to rework for proper model swap than having all in one container.

As always, basic scenario is fine but not fully tested in all scenario.


XAce1337manX (changes requested):
Searching the name of the setting doesn't work. Only "simple blocks" does.
image

XAce1337manX:
Could also remove the TailSphere now that the TailLink model has changed in the ChainContainer but that's a minor thing.
Chain behaviour looks fine otherwise.

KivalEvan:
I may still miss stuff related to option but it's searchable now

As for chain tail sphere, probably better to keep indicator for where the actual tail position is but could revise in the future

PR #605, opened 2025-09-18
Platform lighting bake

KivalEvan:
Live preview lighting now possible when paused.
This also solves ring zoom parity, consequent ring rotation, event skipping, etc.

This obviously uses more memory but whether this can be further optimised for smaller memory footprint is worth looking into.. scary amount of GC.

Have not fully tested on every environment, but you may check if any feels off.

Known issue:
* Chroma lite/gradient event may not properly update affected event on insertion/deletion (honestly, I'm just lazy to deal with this)
* On ring event insertion/deletion at current time may have odd ring spin behaviour
* Ring can go very spinny spin when jumping around timeline or switching from full to static or no lightshow
* When playback from the end, the light does not update properly when it jumps back to beginning

https://github.com/user-attachments/assets/d756fc74-c097-43e8-b712-46a515a22476


Caeden117 (approved):
Code changes look decent. Currently unable to test changes locally until I return home

so uh *hiiiii Bullet*

XAce1337manX:
Looks good. Performance gain is magical.

Two issues:
* The lightshow toggle modes don't indicate which state is currently active. Displaying toggle state would make it obvious that it doesn't change during playback too. 
  image
* If the "Hide Disableable Objects On Load" setting is enabled, the initial lightshow isn't correct for disabled objects when you re-enable the objects and attempting to place an event in the lane for that light throws exceptions
  Can test with https://beatsaver.com/maps/166cf, the back laser rings stay on until the lightshow is repopulated

KivalEvan:
Resolved, probably.

On hindsight, I should've separated out lighting handle and lightshow mode but will deal with them later down the line with platform rework.

XAce1337manX (approved):

PR #606, opened 2025-09-22
Unity 6 Upgrade

KivalEvan:
What does this mean for ChroMapper?
* No mandatory Unity splash screen on start up (perceptibly faster load)
* Loads of regression fix from Unity 2021 that were locked behind including graphic API crash
* Better support for other platform (need to confirm how Apple user is affected)
* And more importantly, better development support

Does this break any plugin from stable?
Existing plugin still work. However, near future development may involve rewriting in certain area where it may or may not break certain, if not all plugins.

Will this be the final Unity build in this development?
Depend if there are any fixes or features that we want by either upgrade or downgrade, 6.0 58f is latest LTS chosen as of Sep 23, 2025

To do:
* Test every functionality (manually) to ensure it work fine in basic scenario
* Check if CI/CD still runs

Known issue:
* Selection box goes nuclear bloom at 0 scale on any axes (temporarily fixed with z epsilon)
* URP post process is now internal, need to find workaround to get full resolution bloom while working on other graphical improvement


XAce1337manX (commented):
Looks good. Will just need to test the build sometime later.
Then see if CI/CD keeps running after merge.

XAce1337manX:
Wait. Wrong target branch. 🙃

PR #607, opened 2025-09-27
Graphical Overhaul Pt.1

KivalEvan:
To properly support current and future content of Beat Saber, a switch from URP to BiRP may be a necessity. This includes bloom behaviour changes. Require thorough exploration and checks. As for now, PR target is to have graphical parity as previous URP. Next part would be to implement Beat Saber-like shader including bloom fog.

What does this bring:
* Alpha bloom and custom tonemapping, similar to Beat Saber for custom content and Vivify compatibility
* Proper alpha blending for lighting
* Updated environment material
* Note and obstacle shader now have dissolve and distortion
* Potential performance improvement
* Possibly fixed giga bloom in certain case
* CJK and extended font support (technically can be separate PRs)
* Organised asset file for shaders and materials (not necessary but I was annoyed at how disorganised it was)
* Upgrade Unity to 6.0 58f2
* Fix on song timeline click not skipping to cursor

To do:
* Support for spot and directional light on custom lit shader, environment lacking spot light and proper directional light to test
* Implement bloom fog (or maybe later PRs), right now and as previous URP, bloom is overcompensating for lack of bloom fog which shouldn't be used this way
* Really need technical people on shader and stuff

Known issue:
* Grid transparency doesn't properly translate alpha transparency (possibly some trick to get around this, otherwise unavoidable)
* Bloom implementation is flawed, need someone with more technical knowledge to look into making proper emissive tonemapping
* Font scaling from extended character set may cause UI to be shifted or disappear
* Resolution scaling does not work above 100%
* Obstacle shader performance issue (due to it having grab pass now), need way to toggle performance and quality mode
* Higher power/usage compared to URP, bloom is expensive to run it seem

https://github.com/user-attachments/assets/6f26bb29-6296-4a29-b331-08341b4fc063


XAce1337manX:
Found some issues outside listed known issues from a quick test.


The buttons on the yes/no dialogue have lost their colours:
image

image


Outline on text needs to be adjusted as it's too thin now:
image

image

Caeden117:
aight lets do this thing

PR #608, opened 2025-09-28
Persist lightshow mode

KivalEvan:
This just makes lightshow mode persist selection during session and fix lightshow showing up even with event load disabled.

Also include fix to modified event type indexing wrong manager causing error.


Caeden117 (approved):
Code LGTM, can test soon™

XAce1337manX:
Test looks good.

PR #610, opened 2025-09-30
Component update optimisation

KivalEvan:
Yet another PR, this time to address unnecessarily expensive component update:
* Measure line and bookmark rendering controller now keep track of active object rather than checking every object and setting valid active.
  * Stops updating when preview mode is on.
* Song timeline controller stop updating when playback in non-normal UI mode, partially because timeline is completely hidden. Text and slider update is expensive even if it's not being shown.
* Since these are time dependent, it is only updated only when time changed in ATSC.

This resulted in component updating up to 3-10x faster.

squash commit thx


Caeden117 (approved):
Code LGTM, will test tomorrow alongside #608

PR #611, opened 2025-10-07
Fix light transition update

KivalEvan:
this is to address transition light not updating if current state is affected by next updated state until state change. this also address chroma light gradient reconstruction but honestly, this is as far as i'd try to reimplement

originally, i wanted this with BiRP PR but the fix also seem to introduce odd behaviour and idk if i've already dealt with them enough


Caeden117 (approved):
LGTM

PR #612, opened 2025-10-18
General Improvement

KivalEvan:
what was originally for small adjustment and improvement ended up with larger scope than intended, couldn't help it but also the codebase is a tangled mess. If there are any part that felt benefit for plugin developer like exposing event or API, do let me know and i'll see if i can do something about it

list of changed stuff if i even remember any:
* Adjusted obstacle transform and visual
* Rework grid rendering and ordering
  * reduced poly count by 5.4k and draw call by idk
  * XY grid now shows precision for precision placement
  * odd grid lane number now possible
* Big adjustment in editor scene object hierarchy
* Restructured codebase partially to their appropriate location
* Renamed some API that may end up break some plugin
  * mainly to address ambiguity and inconsistency
* Fixed and removed unnecessary shader. Goodbye, shader graph
* Fixed light event handling modified conflict and reference error
* Small tweak to prefab like making event text canvas not too large and selection box using cube not more than 12 polygons
* Intersection raycast optimisation
* Rework placement controller
  * faster and less unnecessary work
  * obstacle placement no longer constraint to v2 placement
  * obstacle placement dynamically adjust grid
  * obstacle placement now cancels current placing when switched to other placement
  * event paste to hovered lane on lane hover
* Fix info color scheme not writing alpha

known issue:
* odd number grid lane on center cause all grids to not align properly (thinking of better solution for this since material block is a bit all over the place)
* when grid overlap with another grid (see precision placement grid and event grid), the placement provider fights over intersection causing it to swap between provider many times over depending on position

future plan:
* make Grid and Placement completely event driven
* settle on either static class variable or proper singleton class
  * or look into VContainer or Reflex in the near future for easier time with DI
* dynamically offset grid placement than hardcoding the value (make it possible to have non-whole number offset)


Caeden117 (approved):
LGTM

PR #613, opened 2025-10-19
Bloomfog

Caeden117:
This PR introduces bloomfog rendering to ChroMapper.

- A secondary camera is attached to the active CameraController, rendering only lights into a render texture
- Render texture is downscaled and upscaled with box blurring applied
- A Post Process effect is used to apply the bloomfog texture on top of the active camera, after alpha bloom and before tonemapping

image

Things to maybe consider for this PR:
- Exposing more settings for controlling bloomfog rather than being purely tied to `High Quality Bloom`
- Add bloomfog handling to CMs object shaders (`BloomFog.cginc` should make this relatively straightforward)
- Consider switching from box blur to kawase blur
  - Kawase blur is more performant but box blur is higher quality. Perhaps I turn these into keywords and switch between them with `High Quality Bloom` setting?
- Replace `Graphics.Blit` passes with compute shader bloom
  - Theoretically more performant but it could make the downscale/upscale code more messy


XAce1337manX (commented):
Looks nice. Also lgtm.
Though it does cause merge conflicts with the other PR so we'd probably merge after that one.

PR #614, opened 2025-11-03
Grid and obstacle placement improvement

KivalEvan:
a bit general but it's small PR (finally) to fix obstacle placement grid being only updated when hovered on grid, fix obstacle placement on odd number lane, and some improvement to grid system where it can now offset on decimal without breaking the grid visual


Caeden117 (approved):
LGTM

PR #615, opened 2025-11-04
Fix rotating light not mirroring on same time event

KivalEvan:
small fix, maybe


XAce1337manX:
I don't see a noticeable difference in editor. I'll make a comparison with stable and dev later.

XAce1337manX (approved):

PR #616, opened 2025-11-05
Implement variable NJS on object

KivalEvan:
oh joy, vnjs, this may or may not break noodle stuff but it's a first step

other notable stuff:
* decoupled event action handling from platform and abstracted
* vnjs related stuff decoupled and fixed related test
* arc and chain animation + vnjs adjusted

known issue:
* if vnjs goes very low where hjd is really high, object may not spawn in timely manner
* similar case above, event starts to spawn way too much
* if vnjs goes very high, "despawned" track stretch previous object like arc and obstacle all the way from the back to front

noodle map that uses custom njs may be broken, but i havent seen such problem yet


Caeden117 (approved):
LGTM

PR #618, opened 2025-11-07
Pregenerated noise texture for cutout on objects

UrkiMimi:
this is just a simple change but should optimize some stuff a bit (especially on lower power devices like intel graphics) since the current cutout system is generated through shaders.

other stuff i did
- made distortion obstacles darker and not look like 1.11 menu handles
- added a simple script that sets a global noise texture for cutout
- removed `_WorldScale` for cutout on obstacles for the time being
- did rework on cutout uv mapping to be closer to in game cutout

issues
- texture loops quite badly on large scale objects


Caeden117 (approved):
PR currently looks good to me. If PR is final, can merge

UrkiMimi:
The PR is finalized. You should be good to go with merging it.

PR #621, opened 2025-11-09
Add Alt+Scroll to dot notes

gdt2k:
Probably more of a QOL change but this allows you to use Alt+Scroll on dot notes to make them diagonal
Likely doesn't need testing


Assets/__Scripts/Editor/Input/BeatmapNoteInputController.cs
```diff
@@ -147,6 +147,14 @@ public void UpdateNoteDirection(NoteContainer note, bool shiftForward)
         BeatmapObjectContainerCollection.GetCollectionForType<NoteGridContainer>(ObjectType.Note)
             .RefreshSpecialAngles(note.ObjectData, false, false);
```
Caeden117, line 148: Have you covered/tested the case involving dot notes that are snapped to others? Does behavior work like Beat Saber?

PR #631, opened 2026-02-18
Pack spectrogram values into bytes

Caeden117:
This PR reduces the VRAM requirements of CM's spectrogram by 75%. This is achieved by packing each spectrogram sample into a single `byte` rather than a 4-byte `float`.

This *does* limit the quality resolution of the spectrogram to 256 possible values. However, I am considering this acceptable considering we already do a fair amount of filtering on top of these samples (color gradient, bilinear filtering, etc.) and we've already had a significant number of people who experienced crashing due to the current GPU spectrogram system.

If people do not like 1-byte/8-bit packing, other possible bit-packing possibilities can include:
- 2-byte samples, 50% savings but 65k possible spectrogram values.
- 4-bit samples. 1/8th total size compared to the current float-based spectrogram, but 16 possible values will absolutely cause visible artifacting within the spectrogram.
- 2-bit samples, now you're just being ridiculous.

### DIFF
--- a/Assets/_Graphics/Shaders/Editor/Spectrogram.shader
+++ b/Assets/_Graphics/Shaders/Editor/Spectrogram.shader
@@ -86,12 +86,16 @@ Shader "ChroMapper/Editor/Spectrogram"
 
             float sampleSpectrogram(uint resultIdx)
             {
-                if (resultIdx >= FFTCount)
+                // oops sorry galaxymaster
+                // Fix spectrogram offset by subtracting by our FFT Size multiplied by quality setting
+                uint idx = resultIdx - (FFTSize * FFTQuality);
+
+                if (idx >= FFTCount)
                     return 0.0;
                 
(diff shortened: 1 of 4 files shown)


XAce1337manX:
Quick test with a 25 minute map:
image

Spectrogram looks pretty much identical to me.
image

Assets/_Graphics/Shaders/Editor/Spectrogram.shader
```diff
@@ -86,13 +86,17 @@ Shader "ChroMapper/Editor/Spectrogram"
 
             float sampleSpectrogram(uint resultIdx)
             {
-                // Fix spectrogram offset by subtracting by our FFT Size multiplied by quality setting
-                uint idx = resultIdx - (FFTSize * FFTQuality);
```
XAce1337manX, line 90: @GalaxyMaster2 pointed this out. Spectrogram alignment has shifted a bit.

PR #632, opened 2026-02-27
Beatmap Object Grid Rescale + Model Swap System

KivalEvan:
title. use actual beat saber scale rather than beatwalls unit. this solves couple of issue such as bloomfog and far clip with noodle/chroma, note/obstacle clipping to environment, and overall better scale to actual environment itself

changed:
* rescale beatmap object to beat saber scale
* adjusted Y and Z offset on grid lane and beatmap object
* obstacle lane extend control, similar to note lanes, default to 0

known issue:
* noodle position/scale may not be correct, need to be revised
* certain shader such as wall outline may need to be adjusted
* possibility of rounding error causing placement to be seen as noodle value, haven't done rigorous testing but seems to still be ok

update (3/8/2026):
model swap system is pushed along with this PR

update (3/10/2026):
custom notes support


Caeden117 (approved):
Good to merge following conflict resolution

PR #636, opened 2026-03-24
Edit Mode System + GLS Workflow

KivalEvan:
the GLS lighting is approaching CM

changes:
* added edit mode system (similar to OE with note, GLS and event edit tab)
* added GLS tab & workflow
* object visual highlight when hovered or dragged
* updated text visual on certain grid
* lotsa internal refactor to get here
* selective pasting per edit mode context

known issue:
* no icon for new UI (using text as placeholder)
* box selection end while hover over GLS group enters it instead of ignored
* node editor may not work correctly for GLS event


Caeden117 (approved):
tests pass locally

PR #649, opened 2026-06-22
Rotation Reimplement

KivalEvan:
a bit overdue but it's getting closer

* reimplements rotation event lane back in
* v4 object rotation can be edited
  * `Alt+Q` and `Alt+E` to rotate CCW and CW while hovering object
    * applicable to rotation event
  * `Ctrl+Alt+Q` and `Ctrl+Alt+E` to rotate grid CCW and CW
    * in v4, this applies grid rotation to placement object lane rotation
    * in pre-v4, early rotation event is placed/modified/deleted instead
  * `Alt+R` to copy grid rotation to object
  * `Ctrl+Alt+R` to copy object rotation to grid
  * selectively choose which object type to follow
  * rotation text indicator
* grid and camera follows v4 object rotation during playback
  * use average rotation for object of same time
  * uses editing rotation when not playing
* smoother camera rotation follow
  * reduced motion option available

todo:
* [ ] lane rotation line indicator (similar to game 360 line indicator)
* [ ] idk what else i missed, probably check if i messed up elsewhere without realising


Assets/__Scripts/Beatmap/Base/BaseNote.cs
```diff
@@ -257,6 +258,9 @@ public override int CompareTo(BaseObject other)
             // Compare by angle offset if cut direction
             if (comparison == 0) comparison = AngleOffset.CompareTo(note.AngleOffset);
 
+            // genuinely why is the comment needed?
```
XAce1337manX, line 261: Agree. Honestly should just purge these compare by comments.

XAce1337manX (approved):

PR #657, opened 2026-07-25
ChromaGLS, GLS and node UX, editor state resume per map, new hover input bindings, Chroma UX

EklipZgit:
Review by commit.

Demo here, https://www.youtube.com/watch?v=Xa_Eqe3lQDU

After the demo, easing visualization between GLS nodes was added, and ctrl+shift+scroll and ctrl+alt+shift+scroll bindings added for the remaining properties on Basic Event Ring Rotation and GLS Color Events.

Kind of fixed alt-drag on gls nodes, still doesn't work right but doesn't just warp them around glitchily anymore at least lol.

Fixed easing / HSV preview ribbons, added transition ribbons in GLS, scroll bindings on transition ribbons for easement and hsv, ctrl+shift scroll binding on rotation nodes and lane shifts in GLS


Assets/Input/README.md
Caeden117, line 1: This information may be good to keep for other developers interested in CM's input mechanics, but there seems to be a lot of information related to (what i can only imagine is) your implementation of GLS keybinds. Would recommend removing that contextual information if we are to keep this file.

Assets/__Scripts/UI/Editor/KeybindUpdateUIController.cs
```diff
@@ -12,14 +15,108 @@ public class KeybindUpdateUIController : MonoBehaviour, CMInput.IWorkflowsAction
...
+            Debug.LogError("[KeybindUpdateUIController] eventBoxViewController is null!");
+        }
+        // Deselect the button to prevent it from staying highlighted
+        EventSystem.current?.SetSelectedGameObject(null);
+    }
+
+    private static void AddHotkeyTooltip(Button button, string text, string hotkeyActionMap, string hotkeyActionName)
+    {
+        // Avoid stacking tooltip components if this controller is initialized more than once in the editor.
+        var tooltip = button.GetComponent<Tooltip>() ?? button.gameObject.AddComponent<Tooltip>();
```
Caeden117, line 100: Unity does not properly implement null coalescing operators on MonoBehaviour components. The engine directly overwrites equality operators to check internal engine state, which is not reflected in null coalescence. Please verify that this code isn't susceptible to possible issues (although personally I'd recommend never using null coalesce with Unity objects as a safety precaution)

Assets/__Scripts/Beatmap/Base/BaseEventBoxGroup.cs
Caeden117, line 1: Depending on complexity or frequency of the executed code here, may want to consider moving LINQ usage to ZLinq to limit memory allocation / GC pressure.

Assets/__Scripts/Beatmap/Requirements/ChromaGLSReq.cs
Caeden117, line 1: See previous comment regarding possible ZLinq

GlsLightSchema.jsonc
Caeden117, line 1: Random loose file?

Assets/__Scripts/Actions/Commands/GLSEventRotationCommand.cs
```diff
@@ -28,15 +32,60 @@ public static BaseLightRotationBase SetDirection(BaseLightRotationBase evt, Ligh
...
+        var (newGroup, newEvt) = CopyAndLog(evt);
         newEvt.Loop = value;
         return GLSCommonCommand.TriggerModifyEventAction(
             evt.EventBoxGroupData,
             newGroup,
             newEvt,
             ActionMergeType.ModifyGLSRotationLoop);
     }
+
+    private static void LogMutation(string property, BaseLightRotationBase evt, float value)
```
Caeden117: Consider wrapping in `#if UNITY_EDITOR` or removing altogether so users logs will not be spammed

Assets/__Scripts/Beatmap/Containers/GLSGroupContainer.cs
```diff
@@ -8,31 +10,111 @@ namespace Beatmap.Containers
...
+        private static readonly Stack<GLSGroupContainer> previewGhostPool = new();
+
+        // Aggregate ghost-preview rebuild diagnostics so console output cannot stall routine pool refreshes.
+        private static float nextPreviewDiagnosticTime;
+        private static int previewBuildCount;
+        private static int previewClearCount;
+        private static int previewGhostCount;
+        private static int previewCreatedCount;
+        private static int previewReusedCount;
+        private static int previewSkippedCount;
```
Caeden117: Personally would re-order such that statics are always above instanced variables

Assets/__Scripts/Editor/Grid/Collections/BeatmapObjectContainerCollection.cs
```diff
@@ -681,7 +683,27 @@ public virtual void DeleteObject(
         bool deselect = true,
         bool triggerHandle = true)
     {
-        if (!TryBinarySearch(obj, out var search)) return;
+        // An attached container must point to its exact backing object; otherwise recycle only that orphan visual.
+        if (obj.HasAttachedContainer && LoadedContainers.ContainsKey(obj) && MapObjects.IndexOf(obj) < 0)
```
Caeden117: Curious about the performance of the `IndexOf` check since it always performs a linear search. May cause noticeable delay if `MapObjects` is big enough.

Assets/__Scripts/Editor/Grid/Placements/BoxSelectionPlacement.cs
```diff
@@ -171,23 +192,24 @@ protected override void HandleHitToPlacement(Intersections.IntersectionHit hit,
                 originShove.y -= difference;
             }
 
-            // this math is seriously fucked up man
```
Caeden117: RIP to a real one.

Assets/__Scripts/Editor/Grid/Placements/BoxSelectionPlacement.cs
```diff
@@ -301,10 +308,89 @@ protected override void HandlePlacementToData(PlacementInputState inputState)
...
+        if (Mathf.Approximately(Settings.Instance.GLSOuterTrackGhostNodeOpacity, 0f)) return;
+
+        var epsilon = BeatmapObjectContainerCollection.Epsilon;
+        for (var typeInt = 1; typeInt <= 32; typeInt++)
+        {
+            var type = (ObjectType)(1 << typeInt);
+            if ((selectedTypes & type) == 0) continue;
+
+            var collection = BeatmapObjectContainerCollection.GetCollectionForType(type);
+            if (collection == null) continue;
```
Caeden117: Feels quite jank having to iterate through the `ObjectType` enum then grabbing a `BeatmapObjectContainerCollection`. There is a bunch of junk BOCCs that are not related to GLS - Notes, obstacles, basic events, etc. etc., not even to mention that you *do* have to iterate through all 32 possible bits (even if most of them break early)

Would recommend an alternative here?

Assets/__Scripts/Editor/Grid/Placements/BoxSelectionPlacement.cs
```diff
@@ -301,10 +308,89 @@ protected override void HandlePlacementToData(PlacementInputState inputState)
...
+            {
+                if (!group.HasMatchingTrack(BeatmapObjectContainerCollection.TrackFilterID)
+                    || !IsWithinSelectionXY(GetGlsGroupSelectionPosition(group)))
+                    continue;
+
+                var previewEvent = group.ReadOnlyBoxes
+                    .SelectMany(box => box.ReadOnlyEvents)
+                    .FirstOrDefault(evt => startSongBpmBeat - epsilon < evt.SongBpmTime
+                        && evt.SongBpmTime < endSongBpmBeat + epsilon);
+                if (previewEvent == null || alreadySelected.Contains(group)) continue;
```
Caeden117: See previous comments regarding potential use of ZLinq.

Assets/__Scripts/Editor/Grid/Placements/GLSEventColorPlacement.cs
```diff
@@ -5,10 +5,16 @@
...
     [SerializeField] private BeatmapGLSEventColorInputController inputController;
+    [SerializeField] private ColorPicker colorPicker;
+
+    private int selectedColor;
 
     public override void Start()
     {
         base.Start();
+        // Cross-prefab scene references can deserialize as null; use the shared scene picker as a runtime fallback.
+        colorPicker ??= ColourPicker.ActivePicker ?? FindObjectOfType<ColorPicker>();
```
Caeden117: See previous comment about null coalescence with Unity objects.

Assets/__Scripts/Editor/Grid/CreateEventTypeLabels.cs
```diff
@@ -37,29 +39,32 @@ public void UpdateLabels(EventGridContainer.PropMode propMode, int eventType, in
...
-                catch { }
+                AddLabel(lane++, entries[i].Value.Type, null, entries[i].Value.Name);
+                // Name filters belong to tracks consumed by ring-rotation components, regardless of event-type number.
+                if (!entries[i].Value.Components.HasFlag(BasicEventComponent.RingRotation)) continue;
+
+                var filters = BeatSaberSongContainer.Instance.Map.Events
+                    .Where(x => x.Type == entries[i].Value.Type && !string.IsNullOrEmpty(x.CustomNameFilter))
+                    .Select(x => x.CustomNameFilter)
+                    .Distinct()
+                    .OrderBy(x => x);
```
Caeden117: Potentially expensive operation considering both LINQ *and* the pure fact you are iterating over the entire basic events array (some maps have hundreds of thousands of these).

If current behavior is necessary (seems like it automatically adds lanes for name filtered rings?) then consider ZLinq, or maybe caching CustomNameFilters on map load.

Assets/__Scripts/Editor/Grid/CreateEventTypeLabels.cs
```diff
@@ -37,29 +39,32 @@ public void UpdateLabels(EventGridContainer.PropMode propMode, int eventType, in
 
         if (propMode == EventGridContainer.PropMode.Off)
         {
-            var entries = context.TracksDefinition.Basic.ToList();
+            // Present ordinary light lanes before control lanes without changing their serialized event-type IDs.
+            // Read lane definitions through the dev branch's renamed runtime-context property.
+            var entries = context.TracksDefinition.Basic
+                .OrderBy(entry => entry.Value.Kind == BasicEventKind.Lights ? 0 : 1)
+                .ToList();
```
Caeden117: See ZLinq comments, perhaps `ToList()` can be skipped to ease GC pressure

Assets/__Scripts/Editor/Grid/CreateEventTypeLabels.cs
```diff
@@ -94,6 +125,19 @@ public int EventTypeToLaneId(int eventType)
...
 
+    // Expose the visible basic-event lane mirror so the mirror command can move ordinary light events between displayed lanes.
+    public int MirroredEventType(BaseEvent data)
+    {
+        // Mirror only among visible light lanes; control/event lanes must not become mirror destinations.
+        var lightTypes = laneObjs
+            .Where(entry => context.TracksDefinition.GetBasicOrDefault(entry.type).Kind == BasicEventKind.Lights)
+            .Select(entry => entry.type)
+            .Distinct()
+            .ToList();
```
Caeden117: See previous comments regarding potential use of ZLinq.

Assets/__Scripts/Editor/Grid/CreateEventTypeLabels.cs
```diff
@@ -133,6 +177,17 @@ public int LightIDToLane(int type, int lightID)
...
 
+    // Resolve multi-ID events through displayed physical lanes so hidden IDs cannot become an anchor.
+    public int LightIDsToVisibleLane(int type, IEnumerable<int> lightIDs)
+    {
+        if (lightIDs == null) return -1;
+        return lightIDs
+            .Select(lightID => LightIDToLane(type, lightID))
+            .Where(lane => lane >= 0)
+            .DefaultIfEmpty(-1)
+            .Min();
```
Caeden117: See previous comments regarding potential use of ZLinq.

Assets/__Scripts/UI/Editor/PlacementController/EventBoxViewController.cs
```diff
@@ -96,6 +104,40 @@ private void Start()
...
+            "Reorders all boxes so that X-axis boxes come first, then Y, then Z, making rotation and translation groups easier to read; has no effect on color groups.");
+        AddTooltip(duplicateEventBoxButton,
+            "Dupe",
+            "Creates an exact copy of the currently selected box — including its filter settings and all of its nodes — and inserts it immediately after the original.");
+        AddTooltip(moveUpEventBoxButton,
+            "Move Up",
+            "Shifts the currently selected box one slot earlier in the list, which affects playback order when boxes share overlapping IDs and distributions.");
+        AddTooltip(moveDownEventBoxButton,
+            "Move Down",
+            "Shifts the currently selected box one slot later in the list, which affects playback order when boxes share overlapping IDs and distributions.");
```
Caeden117, line 141: Consider localization for non-English users

Assets/__Scripts/Utils/DataStructures/StateManager.cs
```diff
@@ -84,6 +92,14 @@ protected void HandleInsertUpdateConsequentStateFrom(
 
     protected TState HandleRemoveState(StateChunksContainer<TState, TData> container, TState stateToRemove)
     {
+        // Some environment effects do not cache every matching event; a missing state is already removed.
+        if (stateToRemove == null)
+        {
+            // Keep evidence of cache misses that are safely tolerated during bulk event edits.
+            Debug.LogWarning($"[StateManager] {GetType().Name} skipped removal of an uncached state.");
+            return null;
```
KivalEvan: important to note, it does come at a cost of possible hanging object that is never going to get caught due to false reference until reset, it's preferable we try to catch them early before we add guardrail

Assets/__Scripts/Beatmap/Appearances/EventAppearanceSO.cs
```diff
@@ -188,5 +261,145 @@ public void SetAppearance(
...
+            "easeInOutBack" => "IOB",
+            "easeInElastic" => "InEl",
+            "easeOutElastic" => "OutEl",
+            "easeInOutElastic" => "IOEl",
+            "easeInBounce" => "InBo",
+            "easeOutBounce" => "OutBo",
+            "easeInOutBounce" => "IOBo",
+            "easeStep" => "Step",
+            _ => easing.StartsWith("ease", StringComparison.Ordinal) ? easing[4..] : easing
+        };
```
KivalEvan: we have Easings.cs where all related easing goes to

Assets/__Scripts/Editor/Input/GLSEventHoverMutation.cs
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    public static void MirrorColor(InputAction.CallbackContext context, BaseLightColorBase evt)
+    {
+        if (context.performed && evt != null) GLSEventColorCommand.SetColor(evt, (evt.Color + 1) % 2);
+    }
+
+    public static void AdjustRotation(InputAction.CallbackContext context, BaseLightRotationBase evt, ScrollPrecisionController precision)
+    {
+        if (!context.performed || evt == null || Keyboard.current == null || Keyboard.current.ctrlKey.isPressed || Keyboard.current.shiftKey.isPressed) return;
```
KivalEvan: hardcoding these input key is not advisable, same applies below

Assets/Input/README.md
```diff
@@ -0,0 +1,123 @@
+# ChroMapper Input and Hotkey Architecture
+
+This is the implementation guide for adding or changing ChroMapper hotkeys. The input path spans an authored Unity asset, generated C#, automatic callback installation, overlap filtering, context-specific callbacks, rebind UI, and saved override loading. A change is incomplete until every relevant layer agrees.
+
+## Authoritative Data Flow
+
+1. `Master.inputactions` defines action maps, actions, action types, bindings, composite parts, IDs, and default paths.
+2. Unity's Input System importer generates `Master.cs` with wrapper structs and `I<ActionMapName>Actions` interfaces.
+3. `CMInputCallbackInstaller` creates one shared `CMInput`, discovers generated interfaces, finds scene components implementing them, and installs callbacks.
+4. `InputSystemPatch` prevents a simpler chord from firing when a more specific chord containing the same controls is active. `CMInputCallbackInstaller.Awake` disables Unity shortcut consumption on non-macOS so identical chords can serve context-specific actions in multiple maps.
```
XAce1337manX: Oops, we missed removing the `CMInputCallbackInstaller.Awake` when updating Harmony library to support Apple Silicon. (And I forgot to cherry pick `36c1277863bc8e1037338174af3ac0e7f44f037c` with my previous keybind work.)

Assets/__Scripts/UI/Editor/PlacementController/EventBoxViewController.cs
EklipZgit: Yeah, going to do a pass of tooltips at some point here and was going to figure all that out then, this pr was just scope creeping to infinity and I haven't had time to sit down and try to figure out the tooltip and localization black magic lol.

Edit, actually these tooltips weren't even supposed to go in without me reworking them first, i wasn't sure what this stuff did when i first added them and i never remembered to come back and look them over. Might pull them out of pr for now

Assets/__Scripts/Environments/Effects/Components/TrackLaneRingsManager.cs
```diff
@@ -43,16 +43,14 @@ public void Start()
 
     private void FixedUpdate()
     {
-        var fdt = TimeHelper.FixedDeltaTime;
         var len = Rings.Count;
-        for (var i = 0; i < len; i++) Rings[i].FixedUpdateRing(fdt);
+        for (var i = 0; i < len; i++) Rings[i].FixedUpdateRing(TimeHelper.FixedDeltaTime);
```
Caeden117: Curious if these changes were necessary - typically this is done to cache and avoid repeated variable accesses, especially if the variable is a property (which is effectively a method invocation). This doesnt necessarily matter for Release builds but ChroMapper typically builds in Development mode which doesnt do that kind of optimization.

Assets/__Scripts/Environments/Effects/Components/TrackLaneRingsManager.cs
```diff
@@ -43,16 +43,14 @@ public void Start()
...
     }
 
     private void LateUpdate()
     {
         if (!hasAtsc) return;
-        var intF = TimeHelper.InterpolationFactor;
         var len = Rings.Count;
-        for (var i = 0; i < len; i++) Rings[i].LateUpdateRing(intF);
+        for (var i = 0; i < len; i++) Rings[i].LateUpdateRing(TimeHelper.InterpolationFactor);
```
Caeden117: See previous comment regarding variable caching

Assets/Input/README.md
EklipZgit: Anything actionable for me here? Should i cherry pick that in here or something?

Assets/__Scripts/UI/Editor/Chroma/PaintSelectedObjects.cs
```diff
@@ -25,6 +26,8 @@ public void Paint()
         foreach (var unique in SelectionController.SelectedObjects.DistinctBy(x => x.ObjectType))
             BeatmapObjectContainerCollection.GetCollectionForType(unique.ObjectType).RefreshPool(true);
 
+        FindAnyObjectByType<LightshowController>()?.RefreshLightshow();
```
Caeden117: This call to `FindAnyObjectByType<T>()` should very likely be cached + see previous comments about null coalesce

Assets/__Scripts/UI/Editor/Chroma/StrobeColorPickerController.cs
```diff
@@ -0,0 +1,235 @@
...
+
+    // Let CmData.json restore the strobe toggle whether this flyout has started yet or not.
+    public static void SetLoadedEnabled(bool enabled)
+    {
+        Settings.Instance.PlaceGLSStrobeColor = enabled;
+        Instance?.SetEnabled(enabled);
+    }
+
+    // Reapply the map-scoped setting after UI initialization so the checkbox always matches placement behavior.
+    public static void RefreshLoadedEnabledUi() => Instance?.SyncEnabledUi();
```
Caeden117: See previous comment about null coalesce

Assets/__Scripts/UI/Editor/Chroma/StrobeColorPickerController.cs
Caeden117, line 1: Overall, there is a bunch of initialization / object searching which could be done ahead of time in the Unity Editor. May recommend that approach to save a bunch of expensive Unity calls

Assets/__Scripts/Beatmap/Helper/EnvironmentInfoHelper.cs
```diff
@@ -2,6 +2,8 @@
 
 public static class EnvironmentInfoHelper
 {
+    private const string TheSecondEnvironmentId = "TheSecondEnvironment";
```
Caeden117: leftover debug code?

GlsLightSchema.jsonc
EklipZgit: Eh kind of intended, there's way too many one letter props in the gls lightbox schema and i haven't seen any inline docs anywhere that give people the cliff notes explanation of the props. This was kind of meant to evolve into that. Plus LLMs need stuff like this in context for stuff with arbitrary names they can't infer mappings from, otherwise they'll always be losing context on what the json means as they compact context. So the idea is just having a ref file to @ somewhere in repo to keep that mapping in context easier than @ing several cs files. I can remove it but it's legitimately super effective to flesh one of these out to @ when an llms working with the json schema. I can always pull it out to my outer directory but figured might be useful for others. Could also just throw stuff like this in an /Agents folder or something. Up to you @Caeden117 if you think it's useful somewhere or not

Top-Cat: It definitely doesn't belong directly in the Assets folder.

I would also say this is oddly specific without actually explaining itself as such. Not actual questions but:
What lightshow schema is this valid for?
What type of event group is it?

I also think the documentation on the BSMG wiki is already sufficient https://bsmg.wiki/mapping/map-format/lightshow.html#event-box-groups

Assets/__Scripts/Editor/CmEditorStateData.cs
Caeden117, line 1: While criminally underutilized, we do technically have an existing mechanism in the form of [Editor Metadata](https://github.com/Caeden117/ChroMapper/blob/dev/Assets/__Scripts/Beatmap/Info/Base/BaseInfo.cs#L155-L195). My personal vote is that we should use what already exists but I could be persuaded to use a separate file to store map specific editor settings.

Otherwise, generally not a big fan of how much `FindObject` calls (and derivatives) is being used for a static class. Would almost rather each affected class read/write from editor state themselves rather than having a master class do it

Assets/__Scripts/Beatmap/Appearances/GLSEventCommon.cs
```diff
@@ -95,4 +100,95 @@ public static string GetFloatFXInfo(BaseFxEventFloat evt)
...
+        if (filter == null || sourceGroup == null || groups == null) return false;
+
+        // Order matching nodes by their absolute event time because event ranges from neighboring groups can overlap.
+        followingEvent = groups
+            .Where(group => group.ID == sourceGroup.ID)
+            .SelectMany(group => GetEventsForFilter(group, filter))
+            .Where(evt => evt.JsonTime > source.JsonTime)
+            .OrderBy(evt => evt.JsonTime)
+            .ThenBy(evt => evt.BoxIndex)
+            .FirstOrDefault();
```
Caeden117: Very big LINQ chain here, maybe consider ZLinq depending on how frequently this code is run

Assets/__Scripts/Beatmap/Appearances/GLSEventCommon.cs
```diff
@@ -95,4 +100,95 @@ public static string GetFloatFXInfo(BaseFxEventFloat evt)
...
+        return transition != null;
+    }
+
+    // A group can contain multiple boxes with equivalent cloned filters, so treat their events as one filter sequence.
+    private static IEnumerable<BaseLightColorBase> GetEventsForFilter(
+        BaseLightColorEventBoxGroup group,
+        BaseIndexFilter filter) =>
+        group.Boxes
+            .Where(box => IndexFiltersMatch(filter, box.IndexFilter))
+            .SelectMany(box => box.Events);
```
Caeden117: Ditto ZLinq comment

Assets/__Scripts/Editor/Grid/Collections/BeatmapObjectContainerCollection.cs
EklipZgit: yeah this is a good catch, honestly not sure why this would be necessary over a binary search. I'll try removing it and see what breaks. Definitely going to be related to some ghost-node bug i'm sure, but could be this didn't even fix the issue

GlsLightSchema.jsonc
EklipZgit: removed, fair points

Assets/Input/README.md
EklipZgit: trimmed to the useful information

Assets/__Scripts/UI/Editor/KeybindUpdateUIController.cs
EklipZgit: resolved, and analyzer added to fail the build across the whole codebase. Some ~60 (super ballpark number, didn't count) other locations were also affected, fixed across the board in the final "noisy" commit that just contained zlinq conversions and null prop/coalesce fixes

Assets/__Scripts/Beatmap/Base/BaseEventBoxGroup.cs
EklipZgit: scripted out a dump of all linq usage by line number from my commit diffs and had LLM evaluate all locations, spot checked any it was iffy about, converted all to zlinq that didn't require full array allocations (or had another good justification for why zlinq would be slower). LLM converted most of the non-zlinqable to hard loops as well, so even better. I will resolve all the remaining zlinq comments per this same justification

Assets/__Scripts/Editor/Grid/CreateEventTypeLabels.cs
EklipZgit: done, hard loop, no list

Assets/__Scripts/Editor/Input/GLSEventHoverMutation.cs
EklipZgit: removed, though in at least one spot it still effectively needed this and just refactored out to a common method so at least it is easy to shim mappings in later if necessary. The scroll bindings are intentionally not remappable for now though (before me, already) so no easy way to truly genericize this i don't think

Assets/Input/README.md
EklipZgit: pulled the commit in, removed references to it after it was dropped from the code

Assets/__Scripts/UI/Editor/Chroma/PaintSelectedObjects.cs
EklipZgit: dropped this code completely

Assets/__Scripts/UI/Editor/Chroma/StrobeColorPickerController.cs
EklipZgit: done

Assets/__Scripts/UI/Editor/Chroma/StrobeColorPickerController.cs
EklipZgit: done

Assets/__Scripts/Editor/Grid/CreateEventTypeLabels.cs
EklipZgit: done. Will also be revisiting this later for better support

Assets/__Scripts/Environments/Effects/Components/TrackLaneRingsManager.cs
EklipZgit: reverted, i don't think the change was necessary. Didn't know unity wouldn't optimize this in debug builds, noted

Assets/__Scripts/Beatmap/Helper/EnvironmentInfoHelper.cs
EklipZgit: yep, dropped

Assets/__Scripts/Editor/CmEditorStateData.cs
EklipZgit: Hm, I'd argue we shouldn't be shoving editor meta into the map file itself, just because it's noisy and then people can do something unexpected with it and now you have some undocumented api that changing would break people. I could be persuaded to go the other direction though. Also other tools (OEM editor?) can just strip your stuff out and F us.

WRT to the FindObject, this only happens on map load. Save is all static reference, no Find*'s. It was finicky because the other loading components would overwrite what this did with default state, which is why it does the queued shenanegans. Idk if using serialized fields and making this a game object thingy would affect that though, i'm guessing it would be fine but when I was looking for other FindObjects added by me for earlier comment I did notice this and was hesitant to touch it since it was a pain to get working. up to you.

Assets/__Scripts/Beatmap/Appearances/GLSEventCommon.cs
EklipZgit: This uh, doesn't seem like it should be a linear map scan to begin with. Let me rework this... There's gotta be a better way to store GLS nodes that doesn't require a map scan to find transitions lol. Missed this in my self-audit, whoops

Assets/__Scripts/Utils/DataStructures/StateManager.cs
EklipZgit: All this nonsense should be solved now by ceasing to rebuild the cache in the middle of operations (whoops).

Assets/__Scripts/Editor/CmEditorStateData.cs
Caeden117: Rather than converting this editor state system to a MonoBehaviour, I was just suggesting moving the logic into each specific script that wants persistent editor state. For example, having the GLS placement script explicitly call CmEditorStateData to store, load, and restore placement states, rather than having the state data itself reach out and call everything

EklipZgit: Converted over to what we discussed in discord where components register themselves for load and save

Assets/__Scripts/Beatmap/Appearances/GLSEventCommon.cs
EklipZgit: Fixed, no longer a map scan, maintaining ordered by id group filter and rescanning only the affected range

Assets/__Scripts/Actions/Commands/GLSEventRotationCommand.cs
EklipZgit: removed

Assets/__Scripts/Editor/Grid/Collections/BeatmapObjectContainerCollection.cs
EklipZgit: removed, completely redid this so that the issue it was added for was fixed the proper way

Assets/__Scripts/Editor/Grid/Placements/BoxSelectionPlacement.cs
EklipZgit: redone, just grabbing loaded objects for the 4 gls node types now instead of all objects for all types, good catch.

Assets/__Scripts/Editor/Grid/Collections/BeatmapObjectContainerCollection.cs
```diff
@@ -523,6 +530,63 @@ public abstract class BeatmapObjectContainerCollection<T> : BeatmapObjectContain
 
     public List<T> MapObjects = new();
 
+    // Binary-search the sorted backing list and invoke selection directly without temporary collections.
+    public override void ForEachObjectBetweenSongBpmTime(
+        float start,
+        float end,
+        bool includeOverlappingSliders,
```
Caeden117: For future proofing, may recommend changing this to not be specific to sliders but also walls, since walls can have their end point/duration inside the lower bound, yet starting time outside

Assets/__Scripts/Editor/Grid/Collections/BeatmapObjectContainerCollection.cs
```diff
@@ -523,6 +530,63 @@ public abstract class BeatmapObjectContainerCollection<T> : BeatmapObjectContain
...
+
+        var epsilon = Epsilon;
+        var lowerBeat = start - epsilon;
+        var upperBeat = end + epsilon;
+        var firstIndex = FindFirstObjectAfterSongBpmTime(lowerBeat);
+
+        // Use the bounded loaded window for sliders whose heads precede the lower beat bound.
+        if (includeOverlappingSliders)
+        {
+            foreach (var loadedObject in LoadedContainers.Keys)
```
Caeden117: This is a bit of a code smell? I am unsure why you want to linear scan through loaded containers, especially when the container is using a data structure not meant for fast iteration (`Dictionary<TKey, TValue>`). This method will allocate twice: Once for the allocation of the `KeyCollection`, and again since you are using `foreach`.

Ideally, this should not touch containers at all since you only want to operate with object data. I would recommend moving the slider (ideally slider + wall) code into the real iteration you are doing later.

Assets/__Scripts/Editor/Grid/Collections/BeatmapObjectContainerCollection.cs
```diff
@@ -523,6 +530,63 @@ public abstract class BeatmapObjectContainerCollection<T> : BeatmapObjectContain
...
+
+        for (var index = firstIndex; index < MapObjects.Count; index++)
+        {
+            var obj = MapObjects[index];
+            if (obj.SongBpmTime >= upperBeat)
+                break;
+
+            callback(this, obj);
+        }
+    }
```
Caeden117, line 570: We already have a function that returns a span between two times `Span<T> GetBetween(float, float)`. Additionally, `FindFirstObjectAfterSongBPMTime` could be rewritten to utilize our custom extension method `List<T>.BinarySearchBy` instead of using a manual binary search implementation.

Admittedly, `GetBetween` does not currently consider the slider/wall edge cases where their ending beats can lie in-bounds but starting beats lie outside. Assuming current behaviors dont rely on that edge case *not* being covered, I would recommend fixing slider/wall behaviors at `GetBetween` and simply re-using that method.

Assets/__Scripts/Editor/Grid/Collections/BeatmapObjectContainerCollection.cs
EklipZgit: Dropping the slider code for now, doing a proper fix would require maintaining binary searchable end indexes and keeping it up to date, which sounds like it'd introduce more bugs than this bugfix was worth. Adding a comment documenting the end index approach as a future improvement.

Assets/__Scripts/Editor/Grid/Collections/BeatmapObjectContainerCollection.cs
EklipZgit: dropped my copy, cut over to existing. Dropped the slider stuff, to do it right would require indexing and binary searching event end times and that's complex enough that it'll probably cause new bugs. Left comment with how we can tackle it in future for walls/arcs

Assets/__Scripts/UI/Editor/MirrorSelection.cs
```diff
@@ -23,6 +23,9 @@ public class MirrorSelection : MonoBehaviour
         { (int)NoteCutDirection.Left, (int)NoteCutDirection.Right }
     };
 
+    // Cache the physical mirror domain because standard notes and walls reflect across the complete four-lane beat grid.
+    private static readonly Dictionary<int, int> FullGridLaneMirrorMap = BuildOrderedMirrorMap(new[] { 0, 1, 2, 3 });
```
EklipZgit: not safe assumption, maps can be wider than 4 lanes. Needs to use map lane count

Caeden117 (approved):

KivalEvan:
do need to rebind precision scroll bind as it takes `ctrl+shift+alt+scroll` to 2 separate key bind

not sure why `BeatmapEventInputController` has hardcoded modifier, that should be handled through input action mapping instead

probably have to revise keybind action again in the future, i'm still contemplating about cycle axis, that felt too easily reached for something that's only done once, it's still good to have but not as a scroll

can't say much about rendering QoL or performance so we'll see how that goes

EklipZgit:
> do need to rebind precision scroll bind as it takes `ctrl+shift+alt+scroll` to 2 separate key bind
> 
> not sure why `BeatmapEventInputController` has hardcoded modifier, that should be handled through input action mapping instead
> 
> probably have to revise keybind action again in the future, i'm still contemplating about cycle axis, that felt too easily reached for something that's only done once, it's still good to have but not as a scroll
> 
> can't say much about rendering QoL or performance so we'll see how that goes

Per all the discord discussion, the precision scroll adjust overload stuff will need to be consolidated when all the hover modifier bindings are consolidated, bindings are all good and work great as is and per Bullet we consolidate later. The hardcoded modifiers are necessary for now until we figure out a better system for hover context aware input consumption, which is a more general thing we need to do outside the scope of this PR (and should be easy to retrofit over these changes since everything is abstracted with methods already).

The Axis cycle can be dropped later if we decide it is not useful and something more useful could be bound, but does not make sense to remove it now when it does conflict with anything. 

I fixed 3 additional N^2 cache clears from older code, and brought boost checks from N to LogN. Performance should be pretty good, hopefully. Noticeably improved on high event count maps vs dev branch for me.

Additionally I fixed alt+drag for boost events and GLS groups and GLS events, and fixed all the Boost Event caching bugs :D

Hopefully this covers everything, let me know if you see anything else that needs adjustment~

EklipZgit:
(whoops, meant to resolve Kivals comment, didn't realize I was closing the PR with that comment...)

PR #660, opened 2026-08-09
Dont auto ref roslyn analyzer

EklipZgit:
Dont auto ref rosly analyzer. This file was git ignored since we try not to check in packages, force checking it in to preserve the no-auto-reference that needs to be manually set for unity analyzer dlls apparently


Caeden117 (approved):
bangin pull request my man

Issue #664, opened 2026-08-23
Erroneous word choice in popup


crtbob:
**ChroMapper version**
v0.13.892

**Describe the Issue**
Hello, and thank you for maintaining this software. I created a Github account just to report this, so I apologise if I've done it incorrectly.

I don't think it can be described as a bug, but when certain options are toggled in settings (e.g. Discord Rich Presence) a popup appears that reads:

"ChroMapper may need to be restarted for the changes to take affect."

"affect" is not the correct word to use here; it should instead read "ChroMapper may need to be restarted for the changes to take effect."

**Steps to Reproduce**
1.  Open Chromapper
2.  Click "Options"
3.  Toggle the "Discord Rich Presence" option
4.  Read the popup

**Expected Behavior**
The final word of the popup should be "effect", when it is actually "affect".

EklipZgit:
Fixing this very important bug in next pr ;)

EklipZgit:
I didn't link this ticket properly, @Caeden117 this was resolved by https://github.com/Caeden117/ChroMapper/pull/665

PR #665, opened 2026-08-27
Ring rotation rework and bugfixes

EklipZgit:
Ring rotation interpolated prediction, Hover-scroll ring events in the past to see how they affect the current beat!
Laser speed events same treatment, though have not heavily tested these.
Ring rotation value display improvements.
Various Ring basic event quality of life improvements.
Basic Event Transition chunk reload perf improvements (genericize GLS chunking index).
Fixed a bunch of selection box bugs, implemented int-range track snapping for smooth time-select behavior without allowing selecting outside track bounds.
Fixed issue where loading editor position would have float rounded to be the tiniest epsilon off beat, causing like 1/100 nodes to be considered "ahead" of current beat even when you're on their beat.
Fixed issue where StrobeColor loaded as black.
Plenty of basic event ribbon loading fixes.
Fixed issue where pressing C anywhere would open the Strobe Color picker, instead of just in the GLS views.
Save / resume light id pages as well, the way we resume the right GLS tab if GLS tab was last active during save.
Save and restore the last active GLS tab / light id page so it is the first you go to when returning to that view after a reload.
Added over 250 tests hammering all the corner cases for rotation / zoom / laser speed render caching, ribbon caching behavior, etc.
Fixed auto-save race condition leading to errors (shifted to synchronized context).
Fixed to respect OEM colorSchemes array when no custom colors saved.

8/27 commits:
Fix the very important https://github.com/Caeden117/ChroMapper/issues/664 issue.
Fixed issue where Track Animated rings wouldn't spin, and could appear in the wrong location (by like, a really really large distance, enough to think they're missing entirely). Fixed to match Chroma behavior exactly.
Fix laser speed hover flickering for lockRotation nodes due to invisible ribbon.
Fix Skrillex ring lane names.
Fix Skrillex lane order to make sense.
Fix "HSV" implementation to be Chroma's broken impl, add hidden TrueHSV implementation for when Chroma or ChromaGLS support it that uses existing "good" implementation.
Fixed ability to paste nodes beyond the end of the song, and before beat 0.

8/28:
Fixed an NRE with selection, and use serialized component instead of runtime resolving it.
Fixed The Second Environment's ring zoom step not matching game.

8/30:
Fixed a couple other random init bugs found by the map loader / random settings tester.
Fixed O(n^2) cleanup method


Assets/__Scripts/Editor/Grid/Collections/EventGridContainer.cs
```diff
@@ -339,6 +552,50 @@ private BaseEvent GetNextEventWithSameLightIDOrDefault(BaseEvent e)
...
+        while (lower < upper)
+        {
+            var middle = lower + ((upper - lower) / 2);
+            if (events[middle].JsonTime <= jsonTime)
+                lower = middle + 1;
+            else
+                upper = middle;
+        }
+
+        return lower;
```
Caeden117: Me when manual binary search implementation

Assets/__Scripts/Editor/Grid/Placements/BoxSelectionPlacement.cs
```diff
@@ -655,6 +691,104 @@ private bool IsWithinSelectionXY(Vector2 position) =>
...
+            var middle = lower + ((upper - lower) / 2);
+            if (ranges[middle].x <= rawX)
+            {
+                lower = middle + 1;
+            }
+            else
+            {
+                upper = middle;
+            }
+        }
```
Caeden117: Yet another manual binary search implementation

Assets/__Scripts/Editor/Grid/CreateEventTypeLabels.cs
Caeden117, line 1: Generally unsure if this is worth hardcoding into this script. If Skrillex needs a different order than the rest, I would almost rather that order be defined elsewhere (such as the track definition) so other environments which may need an explicit event order wouldnt need extra code jank

Assets/__Scripts/Editor/AudioTimeSyncController.cs
```diff
@@ -271,7 +271,10 @@ public void LoadEditorState(SimpleJSON.JSONNode data)
     {
         if (data.HasKey("value"))
         {
-            MoveToJsonTime(data["value"].AsFloat);
+            // Info.dat serialization can leave the saved cursor infinitesimally off-grid, so re-lock it to the nearest 1/64 beat on load.
+            // Without this you can get some really weird behaviors where events at the beat you're on don't light up, but other events also on beat do light up (basically some float rounding nonsense).
+            var snappedJsonTime = (float)Math.Round(data["value"].AsFloat * 64f, MidpointRounding.AwayFromZero) / 64f;
+            MoveToJsonTime(snappedJsonTime);
```
Caeden117: Is a 1/64 snapping necessary? would it not make sense to snap to the users defined grid precision?

Assets/__Scripts/Editor/AutoSaveController.cs
Caeden117, line 1: Does this change to saving behavior cause the main thread to potentially lock up during saving? if so, dealbreaker - saving should never hang main thread

Assets/Editor/Environments/Structures/EnvDataInfo.cs
Caeden117, line 1: Ditto previous comment on hardcoding environment-specific behavior

Assets/__Scripts/Editor/AutoSaveController.cs
EklipZgit: nope, it does not

Assets/__Scripts/Editor/AudioTimeSyncController.cs
EklipZgit: great point, had not considered that. Will do

Assets/__Scripts/Editor/Grid/CreateEventTypeLabels.cs
EklipZgit: skrillex definitely needs a different order, it's order is nonsense by default. Will move though

Assets/__Scripts/Editor/Grid/Placements/BoxSelectionPlacement.cs
EklipZgit: sure, i didn't see one that accepted vectors though, and its also binary searching ranges. I can't really make vector implement an interface lol

EklipZgit: i guess i could make a whole class/struct for containing it and make that adhere to some interface? Sounds pretty convoluted though imo, esp when there's nuances to the end of this binary search that behave different than the other methods

Caeden117: not only a `ranges.BinarySearchBy(rawX, range => range.x);`? assuming these lists are sorted by X (which i imagine they must be if theres binary search to begin with)

i do also agree that a whole data structure to adhere to the de-duplication doesnt make as much sense. if my idea above cannot work out, or if the additional processing underneath makes `BinarySearchBy` unsafe to use, this manual implementation can stay (just document the justification before someone else tries to de-dupe)

Assets/__Scripts/Editor/Grid/Collections/EventGridContainer.cs
EklipZgit: ok this one I missed lol, definitely should not be reimpl'd here

Assets/__Scripts/Editor/Grid/Placements/BoxSelectionPlacement.cs
EklipZgit: That's fair and should work. I hadn't considered looking for something like that on the premise that range => range.x makes a direct field access into a func call on the stack which is much less efficient than direct field access, but I guess logarithmic stack calls isn't a big deal, and thats assuming the compiler doesn't have magic to optimize out the func anyway which for all I know it might. And this is just running once per mouse position update I think (or once per frame) so not like it needs to be absurdly efficient anyway. I'll change it

Assets/Editor/Environments/Structures/EnvDataInfo.cs
EklipZgit: this is necessary, skrillex has lanes named wrong (in-game calls ring-rotation-2 Laser Speed (or rather, ring rotation 1 because they're in the wrong order too lol)). And, this is at build-environment editor generation time, rather than runtime, no? So isn't this exactly where you DO want it? This is where I would move the other comments code to, pretty sure

_This is exactly where Kival said to put it lol_

Assets/Input/README.md
```diff
@@ -75,3 +75,21 @@ Test the default binding, rebind it, save and reload it, then rebind the restore
 - Test keybind discovery, rebinding, saved override reload, and rebinding after reload.
 - Confirm generated `Master.cs` matches `Master.inputactions`.
 - Run a clean Unity compile/build.
+
+
+
+
+## NOTES 21f1
+
+# m_ScrollDeltaBehavior: 0
```
EklipZgit, line 84: these new comments should be removed, missed this

Assets/Editor/Environments/Structures/EnvDataInfo.cs
KivalEvan: i don't oppose to this, this is where any postfix for environment data should be done instead of during runtime

that said, from [Assets/__Scripts/Editor/Grid/CreateEventTypeLabels.cs](https://github.com/Caeden117/ChroMapper/pull/665/files/d4d2b7c89b535a247c075dc9f6f85bf794329fac#diff-5c0adb13e5edfdb2ebeb9f46360e8f2901ec146c695deae9d62f0588d5a250cb) comment earlier, the event lane reordering should be handled here, technically can also be manually re-ordered in TrackDefinitionSO itself

could have better way to do remapping but works for now

Assets/Editor/Environments/README.md
```diff
@@ -0,0 +1,56 @@
+# ChroMapper environment generation
+
+The JSON files under `Assets/__Scenes/Environments/Data/` are the UGEcko game-export source files consumed by the current environment generator. In this document, "export JSON" means one of those checked-in files, such as `Assets/__Scenes/Environments/Data/TheSecondEnvironment.json`; it does not mean a beatmap JSON file. Keep these exports faithful to the game dump. ChroMapper-specific corrections belong in the import path, not in those source files.
```
EklipZgit, line 3: this readme is useful for documenting how the EnvData generation process works, and where responsibility for correcting things lies, but it should not document things like the Skrillex specific fixes that are better simply getting a 1 sentence mention of that .cs file being the correct place to put env corrective hacks in (since it is not Ugeckos responsibility to correct bad game decomp names in their output, and this is where we process that data into CM scenes).

The rest of this LLM nonsense should be stripped out so this is just useful overview of responsibilities and such.

Assets/Editor/Environments/Structures/EnvDataInfo.cs
EklipZgit: yeah. I audited every other environment for mis-named lanes and Skrillex is literally the only one. Beat Games must have just missed it when they cloned another env 🤷

EklipZgit: so, while maybe lane orders could use some clean up, this should be the only lane name hacks necessary for all the existing envs so far, so no sense trying to overcomplicate it

Assets/__Scripts/Editor/Grid/CreateEventTypeLabels.cs
EklipZgit: This code should move to EnvDataInfo.cs

Assets/__Scripts/Editor/AutoSaveController.cs
EklipZgit: hm, apparently it can block the main thread. I've had this code in for A WHILE and have noticed zero freezing during live playback during autosaves, but AI says this does force it to run on the main thread. I'll revert it, it was just to prevent the random race condition errors that users confusingly get in the console. I guess better fix then would be to just add a bit more info to the error saying "if you aren't getting this every auto-save then you can safely ignore this" or something so it looks less alarming when it happens. I don't have the original error though so i'll have to wait till I see it again to put something like that in place, so for now I'll just revert this

Assets/__Scripts/Environments/Effects/Basic/TrackLaneRingsPositionEffect.cs
```diff
@@ -22,10 +24,31 @@ private void Awake()
...
+            var spawners = transform.root.GetComponentsInChildren<TrackLaneRingsPositionSpawner>(true);
+            for (var spawnerIndex = 0; spawnerIndex < spawners.Length; spawnerIndex++)
+            {
+                if (spawners[spawnerIndex].EffectManager == this)
+                {
+                    Visual = spawners[spawnerIndex];
+                    break;
+                }
+            }
+        }
```
Caeden117, line 40: Curious if this is worth a fix that is applicable to all environments (again harping on the idea of having less environment-specific fixes/code) - if not feasible, no worries here

Assets/Editor/Environments/Structures/EnvDataInfo.cs
```diff
@@ -172,14 +213,41 @@ private static string GetBasicTrackName(string environmentId, int eventType, str
...
+        string environmentId,
+        List<TrackDefinitionBasic> tracks)
+    {
+        // BillieTrackDefinitionImportUsesCorrectedLaneOrder selects a cached, environment-specific order only at generation time.
+        var presentationOrder = environmentId switch
+        {
+            BillieEnvironmentId => BillieBasicEventOrder,
+            SkrillexEnvironmentId => SkrillexBasicEventOrder,
+            _ => null
+        };
```
Caeden117: This is more nitpicky than an overall design change, but may be worth switching this to use a `Dictionary<string, Dictionary<int. int>>` so new environment remaps are simply a new entry into the outer dict, rather than having multiple places to change

Caeden117 (approved):

Assets/__Scripts/Environments/Effects/Basic/TrackLaneRingsPositionEffect.cs
EklipZgit: I think this is an all-environments fix, it was just that only GreenDayGrenade ended up initializing things in the order that reproduced it. Its possible maybe something more core could have it's order shifted around so that this hack isn't necessary though. I'll see if I can do that with the V4 conversion PR that I'll have out pretty shortly after this merges, I'll have time while testing that to iterate on this as well

Assets/Editor/Environments/Structures/EnvDataInfo.cs
EklipZgit: agreed, working on it

EklipZgit: done

Assets/__Scripts/Environments/Effects/Basic/TrackLaneRingsPositionEffect.cs
EklipZgit: working in this in other PR actively

PR #667, opened 2026-08-31
CI test fixes for input and audio, runtime->init perf optimizations, serialize Visual, docs, beat rounding drift correction

EklipZgit:
CI test fixes for input and audio.

Rotation runtime->init perf optimizations, including serializing Visual to avoid runtime checks.

Docs around CI input and audio limitations.

Beat rounding drift correction (basically snap to grid when moving something within .001 (or whatever your decimal precision is currently set to) of the grid. Prevents shift-moving events from having float drift slowly accumulate until your lights are "after" the current beat by 0.001 and you get bad previews of current time, etc.

Test-ChroMapper.ps1 script that runs the unity tests via CLI (only works when Unity does not have the project open) and outputs any failures logs to CLI. Should(?) work on linux if pwsh installed, but did not test on linux


Assets/Tests/Test-ChroMapper.ps1
Caeden117, line 1: Was this meant to be checked in?

EklipZgit: yes, figured could be useful to others and should work on linux as well provided pwsh is installed

Top-Cat: Is it AI that has started convincing people that powershell is **the** cross platform scripting tool? (Is this why MS invested in OpenAI)

I've seen this a couple times now when for years it's been shell scripts...

EklipZgit: No, pwsh is my scripting language of choice for anything like this, and it is cross platform. I was probably top 200 most knowledgeable powershell users in the world around 8 years ago and have worked with it down to the AST level, so perhaps I'm slightly biased but imo object oriented scripting is vastly superior to sh. I'm sure I wont convince you, though ;)

I can delete it, I don't really care, just figured it might be useful lol. Caeden mentioned checking in local build scripts if I wanted like 2 months ago, this was the first one that seemed like something you didn't already have

EklipZgit: https://dev.to/crt0r/a-linux-geek-tried-powershell-it-feels-powerful-indeed-1jl4
this would be why you're seeing it more, if you care

EklipZgit: @Caeden117 @Top-Cat was there something you wanted me to do here? Should I just mark this resolved...? Remove it? What's the play

Caeden117: I dont mind the test script staying checked in, maybe move it to the tests folder? The only thing I am unsure about is whether or not it has to stay in the root of the repo

EklipZgit: sure, will do

EklipZgit: done

Caeden117 (approved):

PR #668, opened 2026-09-03
Vnjs and v4 walls in v3, and gls polish, and ChromaGLS + BeatToTheFuture

EklipZgit:
Allow placing VNJS in V3 maps (prompt for BeatToTheFuture requirement on first placement).
Support BeatToTheFutures mappedForOldRingPropagationSpeed customData field to detune ring prop rate by 40% to match old game version behavior (so people who tuned their maps against 1.40.8).
Auto-save new songs instead of having that weird "make sure to click save" prompt. We already made the folder why would we not make the .dat.
Add GLS light color event hover keybind for toggling strobe fade (Shift+Scroll).
Add VNJS easing cycling bind (ctrl+shift+scroll, same as all other easing hovers).
Condensed Environments build info readme.
Properly support StrobeColor HDR for ChromaGLS.
Auto-convert prompt from Mapping Extensions to Noodle (probably doesn't trigger everywhere).
Pseudo XYZ lanes visible but grayed out when opening rotation and translation GLS Event Groups, so you can place XYZ without weird lane buttons. (Pseudo lanes do not serialize unless something is placed in them. Once something is placed in them or their filters are modified they become real lanes permanently whether or not something is in them).
Added "YEET" Hover keybind, (Hover)+Shift+Z to YEET or unYEET a GLS translation event.
+IDs button clones current axis events out to id lanes instead of deleting.
+Axes buttons just makes the ghost axes permanent and no longer deletes all your events.
+Axes IDs button does the combination of both ^
Some random minor GLS bugfixes


EklipZgit:
missed a commits excessive ai comments which probably means I didn't audit that commit properly, will investigate in the morning and resubmit when all's good

PR #670, opened 2026-09-06
Port improved chain placement behavior from slinkmapper

Vainstains:
(pretty much all just workflow QOL)
Change chain placement to be more robust, following the implementation from my SlinkMapper plugin for stable.
- pairs are flipped to make the head first
- if the head is very close to the tail time-wise, flip based on direction and position dot product comparison to make the first note in the swing the head
- if the head is a dot, and the tail is an arrow, estimate a sensible angle for the head based on the tail and make a dot chain

### DIFF
--- a/Assets/__Scripts/Editor/Grid/CommonBeatmapUtils.cs
+++ b/Assets/__Scripts/Editor/Grid/CommonBeatmapUtils.cs
@@ -1,0 +1,104 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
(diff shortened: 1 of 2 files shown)


Caeden117 (commented):
Code itself looks good to me. There are a bunch of helper methods that aren't necessarily specific to chains, so I would consider moving these to a more public and generic location but this isn't breaking to the PR.

Caeden117 (approved):

